PR #16, opened 2024-11-25
Allow using AdapterBase in MethodMock

piotro888:
Reopened rebased #7 

---

I accidentally found new use for @def_method_mock (?), because it worked on older Transactron in other project.

It is nice to use (and more intuitive than in testbench example use) for proving input / collecting output of input-only and output-only Methods in modules, instead of writing call loops, and can be used directly for driving SimpleTestCircuit generated TestbenchIOs.

### DIFF
--- a/test/testing/test_method_mock.py
+++ b/test/testing/test_method_mock.py
@@ -1,0 +1,47 @@
+import random
+from amaranth import *
+from amaranth.sim import *
+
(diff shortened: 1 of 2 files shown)


tilk:
The example test feels kind of weird. The most troubling thing is the active process: if forgotten, the test will be passing without testing anything.

PR #20, opened 2024-11-27
Lib testing refactor

tilk:
This PR splits the giant `test_transaction_lib` file into a number of smaller names, with names corresponding to the names of source files in the `lib` directory.


piotro888 (approved):
There is existing `test/test_connectors.py` file with `TestStableSelectingNetwork`, it should be merged. Otherwise LGTM

tilk:
> There is existing `test/test_connectors.py` file with `TestStableSelectingNetwork`, it should be merged. Otherwise LGTM

I know about it - but I think that, for consistency, `StableSelectingNetwork` needs to be moved out of `lib` (and probably made into a `Component`), as it doesn't depend on Transactron core.

PR #21, opened 2024-11-27
Remove parameterized

tilk:
This PR removes the dependency on `parameterized`, `pytest.mark.parametrize` is used instead. Based on #20.

### DIFF
--- a/test/utils/test_utils.py
+++ b/test/utils/test_utils.py
@@ -1,5 +1,6 @@
 import unittest
 import random
+import pytest
 
 from amaranth import *
 from transactron.testing import *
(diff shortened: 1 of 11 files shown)


test/utils/test_utils.py
```diff
@@ -112,17 +109,15 @@ def elaborate(self, platform):
...
-    size: int
-
-    def setup_method(self):
+    @pytest.fixture(scope="function", autouse=True)
+    def setup_fixture(self, size):
+        self.size = size
         random.seed(14)
         self.test_number = 40
         self.m = CLZTestCircuit(self.size)
+        yield
```
piotro888, line 120: Why there is `yield` added here and not in next test?

piotro888 (approved):
I didn't know pytest parametrize integrates that well, nice

test/utils/test_utils.py
tilk: It turns out it's not needed, when a fixture is not a generator it's called before test runs, which is what we want here. Will remove.

PR #23, opened 2024-11-30
Wide FIFO

tilk:
This PR introduces a wide FIFO data structure, which allows writing and reading multiple entries in a cycle. The entries are left-aligned and the order is preserved.

TODO:
* [x] Merge #22
* [x] Add documentation
* [x] Test the barrel shifter module


transactron/lib/fifo.py
```diff
@@ -114,6 +118,176 @@ def _() -> None:
...
+
+        max_width = max(self.read_width, self.write_width)
+
+        storage = [memory.Memory(shape=self.shape, depth=self.depth, init=[]) for _ in range(max_width)]
+
+        for i, mem in enumerate(storage):
+            m.submodules[f"storage{i}"] = mem
+
+        write_ports = [mem.write_port() for mem in storage]
+        read_ports = [mem.read_port(domain="sync", transparent_for=[port]) for mem, port in zip(storage, write_ports)]
```
lekcyjna123, line 197: Is the transparency needed?

transactron/lib/fifo.py
```diff
@@ -114,6 +118,176 @@ def _() -> None:
...
+            m.d.comb += ens.eq(Cat(i < count for i in range(max_width)))
+            m.d.comb += Cat(port.en for port in write_ports).eq(rotate_left(ens, write_col))
+            m.d.av_comb += [write_ports[i].data.eq(shifted_data[i]) for i in range(max_width)]
+            m.d.comb += write_count.eq(count)
+            m.d.sync += incr_row_col(write_row, write_col, write_row, write_col, incr_write_row, count)
+
+        m.d.comb += next_read_row.eq(read_row)
+        m.d.comb += next_read_col.eq(read_col)
+        m.d.sync += read_row.eq(next_read_row)
+        m.d.sync += read_col.eq(next_read_col)
```
lekcyjna123, line 268: High surprise factor. A comment in the code will be useful.

test/utils/test_shifter.py
```diff
@@ -0,0 +1,144 @@
...
+        ],
+    )
+    @pytest.mark.parametrize(
+        "shift_fun, shift_kwargs, test_fun",
+        [
+            (shift_vec_left, lambda mkc: [], lambda val, offset, mkc: [mkc(0)] * offset + val[: len(val) - offset]),
+            (shift_vec_right, lambda mkc: [], lambda val, offset, mkc: val[offset:] + [mkc(0)] * offset),
+            (
+                shift_vec_left,
+                lambda mkc: [("placeholder", mkc(0))],
```
lekcyjna123, line 110: Shouldn't be `mkc(1)`? With `0` test is nearly the same as previous `shift_vec_left`.

test/lib/test_fifo.py
```diff
@@ -62,3 +68,57 @@ async def target(sim: TestbenchContext):
...
+class TestWideFifo(TestCaseWithSimulator):
+    async def source(self, sim: TestbenchContext):
+        cycles = 100
+
+        for _ in range(cycles):
+            await self.random_wait_geom(sim, 0.5)
+            count = random.randint(1, self.write_width)
+            data = [const_of(random.randrange(2**self.bits), self.shape) for _ in range(self.write_width)]
+            await self.circ.write.call(sim, count=count, data=data)
+            await sim.delay(2e-9)
```
lekcyjna123, line 82: Usage of `2e-9` is not very clear to me. Why that value and not `1e-9`? Shouldn't we use some variable with cycle length instead?

lekcyjna123 (approved):

transactron/lib/fifo.py
tilk: I believe so, same as in `BasicFifo`. Explanation: if the FIFO is empty and a `write` is performed in cycle 1, then if `read` is called in cycle 2, it needs to return the value written in cycle 1, which is provided thanks to transparency. If you remove transparency, the test fails.

test/lib/test_fifo.py
tilk: Because 1e-9 is used in a different testbench process. This is used to enforce execution order without synchronization primitives, which sadly are not provided (?) by Amaranth.

test/utils/test_shifter.py
tilk: Will fix.

PR #28, opened 2024-12-12
Order preserving allocator

tilk:
This PR introduces an allocator which preserves allocation order information in the form of identifier permutation. The implementation performs deallocation by rotating some continuous range in the permutation.

### DIFF
--- a/transactron/lib/allocators.py
+++ b/transactron/lib/allocators.py
@@ -63,7 +63,7 @@ class PriorityEncoderAllocator(Elaboratable):
         return m
 
 
-class LRUOrderAllocator(Elaboratable):
+class PreservedOrderAllocator(Elaboratable):
     """Allocator with allocation order information.
 
     This module allows to allocate and deallocate identifiers from a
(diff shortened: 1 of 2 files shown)


transactron/lib/allocators.py
```diff
@@ -60,3 +61,74 @@ def _(_, ident):
...
+            return {"ident": order[used]}
+
+        @def_method(m, self.free_idx)
+        def _(idx):
+            for i in range(self.entries - 1):
+                with m.If((i >= idx) & (i + 1 < incr_used)):
+                    m.d.sync += order[i].eq(order[i + 1])
+            for i in range(self.entries):
+                with m.If(i + 1 == incr_used):
+                    m.d.sync += order[i].eq(order[idx])
```
piotro888, line 120: Not very critical, but I think it can be cheaper in hardware by always shifting `[idx:]` suffix and putting freed identifier at end of `order` (avoids second mux). Additionally it would produce more balanced identifier use.
Would it break some intended use cases, or is it worth adding?

tilk: As I consider free IDs to be undistinguishable, the solution you present is indeed better. Done.

piotro888 (approved):

PR #29, opened 2024-12-17
CLZ and CTZ for non power of two inputs

Hazardu:
In this PR i add zeros to input signals of `count_leading_zeros` to extend the shape to nearest power of two

Original issue:
https://github.com/kuznia-rdzeni/coreblocks/issues/751

### DIFF
--- a/test/utils/test_utils.py
+++ b/test/utils/test_utils.py
@@ -11,6 +11,7 @@ from transactron.utils import (
     count_leading_zeros,
     count_trailing_zeros,
 )
+from amaranth.utils import ceil_log2
 
 
 class TestAlignToPowerOfTwo(unittest.TestCase):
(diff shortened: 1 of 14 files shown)


transactron/utils/amaranth_ext/functions.py
```diff
@@ -53,28 +53,24 @@ def iter(s: Value, step: int) -> Value:
...
-
-    value = iter(s, xlen_log)
+    slen = len(s)
+    slen_log = ceil_log2(slen)
+    closest_pow_2_of_s = 2 ** slen_log
+    zeros_prepend_count = closest_pow_2_of_s - slen
+    if zeros_prepend_count > 0:
+        value = iter(Cat(C(0, shape=zeros_prepend_count), s), slen_log) 
+    else:
+        value = iter(s, slen_log)
```
tilk, line 63: I think the `if` is unnecessary and the `else` case can be removed.

tilk (commented):
There is a merge conflict.

test/utils/test_utils.py
```diff
@@ -109,7 +110,7 @@ def elaborate(self, platform):
         return m
 
 
-@pytest.mark.parametrize("size", range(1, 7))
+@pytest.mark.parametrize("size", range(1, 128))
```
tilk, line 113: Reduce the number of tests.

test/utils/test_utils.py
```diff
@@ -152,7 +154,7 @@ def elaborate(self, platform):
...
```
Arusekk, line 157: Notes from discussion:

```suggestion
@pytest.mark.parametrize("size", [1,2,3,4,5,6,7,8,9,63,64,65,97,98,127,128])
```
(less tests sample)

Arusekk (approved):

PR #33, opened 2024-12-19
Improved SimpleTestCircuit

tilk:
This PR changes `SimpleTestCircuit` so that `Adapter`s and `TestbenchIO`s can be automatically constructed for required methods. Type annotations is used to distinguish required methods from provided ones. Based on #30.

### DIFF
--- a/test/testing/test_method_mock.py
+++ b/test/testing/test_method_mock.py
@@ -10,6 +10,44 @@ from transactron.testing.method_mock import MethodMock, def_method_mock
 from transactron.lib import *
 
 
+class SimpleMethodMockTestCircuit(Elaboratable):
+    method: Required[Method]
+    wrapper: Provided[Method]
+
+    def __init__(self, width: int):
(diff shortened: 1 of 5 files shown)


lekcyjna123 (approved):

piotro888:
Great!
Should we now shift from accepting external `Methods` in `__init__` to using `Required` + connectors everywhere?

piotro888:
related issue: #12

tilk:
> Great! Should we now shift from accepting external `Methods` in `__init__` to using `Required` + connectors everywhere?

I don't see how connectors fit into this, could you elaborate? My original intention was to use `proxy` to satisfy dependencies, like this:

```python
c1 = ProvidesSomething()
c2 = RequiresSomething()
c2.required.proxy(c1.provided)
```

(If this use case becomes standard, then `proxy` is a weird name and renaming is probably in order.)

I don't know if it is worth it to do a massive refactor now (but maybe it is?). There are also places where having method arguments seems useful (method transformers?). For now, let's try to use this and see where it takes us.

PR #36, opened 2025-01-06
Adapt transactron.lib.storage to amaranth.lib.data

piotro888:
* Previously memory element layout was `LayoutList` that was latter flattened to unsigned shape. Now memory accepts `ShapeLike` shape that is forwarded to memory
* granularity support was accidentally removed, fixed that (supported only `unsigned` and `ArrayLayout` shapes asserted by amaranth)
* renamed `elem_count` to `depth` for consistency with Amaranth

### DIFF
--- a/transactron/lib/storage.py
+++ b/transactron/lib/storage.py
@@ -56,7 +56,7 @@ class MemoryBank(Elaboratable):
         granularity: Optional[int]
             Granularity of write. If `None` the whole structure is always saved at once.
             If not, shape is split into `granularity` parts, which can be saved independently (according to
-            `amaranth.lib.mem` granularity logic).
+            `amaranth.lib.memory` granularity logic).
         transparent: bool
             Read port transparency, false by default. When a read port is transparent, if a given memory address
             is read and written in the same clock cycle, the read returns the written value instead of the value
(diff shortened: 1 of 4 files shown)


transactron/lib/storage.py
```diff
@@ -254,24 +265,25 @@ class AsyncMemoryBank(Elaboratable):
...
             The format of structures stored in the Memory.
-        elem_count: int
+        depth: int
             Number of elements stored in Memory.
         granularity: Optional[int]
-            Granularity of write, forwarded to Amaranth. If `None` the whole structure is always saved at once.
-            If not, the width of `data_layout` is split into `granularity` parts, which can be saved independently.
+            Granularity of write. If `None` the whole structure is always saved at once.
+            If not, shape is split into `granularity` parts, which can be saved independently (according to
+            `amaranth.lib.mememory` granularity logic).
```
tilk, line 286:
```suggestion
            `amaranth.lib.memory` granularity logic).
```

tilk (approved):

transactron/lib/storage.py
```diff
@@ -114,7 +123,9 @@ def _(i: int):
                 m.d.sync += overflow_valid[i].eq(0)
             with m.Else():
                 m.d.sync += read_output_valid[i].eq(0)
-            return Mux(overflow_valid[i], overflow_data[i], read_port[i].data)
+
+            # Amaranth Mux drops lib.data Layouts
+            return {"data": View(self.shape, Mux(overflow_valid[i], overflow_data[i], read_port[i].data))}
```
tilk, line 128: Unfortunately this is a bug. If `self.shape` is not a data layout, this fails.

PR #37, opened 2025-01-06
Add `until_all_done` to `CallTrigger`

piotro888:


Arusekk (approved):
Maybe worth some short docstring a la `Same as until_done but wait for all results instead of any result.`

tilk:
Is this useful? Waiting for all calls effectively means ignoring some call successes, which is typically not what you want if any of them perform side effects.

tilk:
> Maybe worth some short docstring a la `Same as until_done but wait for all results instead of any result.`

@piotro888, could you do it?

PR #41, opened 2025-01-17
Fix memory banks for non-data shapes and test

tilk:
This PR fixes a bug introduced in #36 and modifies tests to verify the fix.


tilk:
Bypass merging as this is fairly obvious and fixes a blocking bug.

PR #42, opened 2025-01-18
`ComponentInterface` typing wrapper for lib.wiring interfaces

piotro888:
Redesign of concept from Coreblocks PR https://github.com/kuznia-rdzeni/coreblocks/pull/726.
Class that can be used both for typing and to get amaranth Signature for use in Component constructor.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "transactron"
 dynamic = ["version"]
 dependencies = [
     "amaranth == 0.5.3",
-    "amaranth-stubs @ git+https://github.com/piotro888/amaranth-stubs.git@e25a7fa11d4a0d66ed18190f31b60914f222e74c",
+    "amaranth-stubs @ git+https://github.com/kuznia-rdzeni/amaranth-stubs.git@8841b43f0a1e8aa0453dc0c8df1dc38818a8fcac",
     "dataclasses-json == 0.6.3",
     "tabulate == 0.9.0"
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "transactron"
 dynamic = ["version"]
 dependencies = [
     "amaranth == 0.5.3",
-    "amaranth-stubs @ git+https://github.com/kuznia-rdzeni/amaranth-stubs.git@d62b68241fa1881244470e63931ceb635ac91044",
+    "amaranth-stubs @ git+https://github.com/piotro888/amaranth-stubs.git@e25a7fa11d4a0d66ed18190f31b60914f222e74c",
     "dataclasses-json == 0.6.3",
     "tabulate == 0.9.0"
 ]
--- a/test/utils/test_component_interface.py
+++ b/test/utils/test_component_interface.py
@@ -1,6 +1,7 @@
 from unittest import TestCase
 
-from amaranth import Signal
+from amaranth import Signal, unsigned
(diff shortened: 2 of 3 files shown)


piotro888:
Improvement from previous version: now separate `FlippedComponentInterface` is used to implement flips cleanly, without using `copy` and getting `is_flipped` under try/catch.

piotro888:
amaranth-stubs commit needs to be changed before merge

tilk:
Can this handle arbitrary `ShapeLike`s as members?

piotro888:
Yes, tested here for `int`:
https://github.com/kuznia-rdzeni/transactron/blob/81cf8731acf855211c29d206f71c5abdb5693668/test/utils/test_component_interface.py#L21
and is `ShapeLike` in type stubs.

I can additionally include some `ShapeCastable` field in testbench.

piotro888:
Turns out there was an unnecessary early Shape.cast that broken View wrapping trick around Signal, and Layouts were flattened. Fixed now and added to tests.

tilk:
This `ComponentSignal` feels hacky, which could be fragile with future Amaranth updates. Is it possible to implement it as a `ShapeCastable`?

piotro888:
There is a problem with typing of `ComponentSignal.shape()`. It is required from View that it returns `Shape`, but it is not true from lib.data fields that use a hack to wrap around `Signal` and return `ShapeLike`/`ShapeCastable`? instead.

`AbstractSignal` with shape changed to always return `Shape | ShapeCastable` (without Value inheritance) should solve this issue, but loosen typing a bit, but I don't think that Shape ShapeCastable distinction is that critical in calling shape() and I don't see other way to do it.

piotro888:
I think this problem is not addressed in current `amaranth-stubs`: `Signal(ArrayLayout(2, 3)).shape() == ArrayLayout(2, 2))`, but according to stubs it must be `Shape`.
It is true for Amaranth Signal, but not one extended with lib.data

tilk:
> I think this problem is not addressed in current amaranth-stubs: Signal(ArrayLayout(2, 3)).shape() == ArrayLayout(2, 2)), but according to stubs it must be Shape.

This is not true. `Signal(ArrayLayout(2, 3))` is not a `Signal`, it's a `View`.

piotro888:
> This is not true. Signal(ArrayLayout(2, 3)) is not a Signal, it's a View.

How this is described in stubs? Maybe similar method could be used here?
I know about Signal -> View conversion, but didn't expect it to be possible to define in stubs.

tilk:
> > This is not true. Signal(ArrayLayout(2, 3)) is not a Signal, it's a View.
> 
> How this is described in stubs?

See `_SignalMeta`: https://github.com/kuznia-rdzeni/amaranth-stubs/blob/d62b68241fa1881244470e63931ceb635ac91044/amaranth-stubs/hdl/_ast.pyi#L497

(By the way, there is a slight problem in the stubs here - the overload for `ShapeLike` should be for `ShapeLike`s which are not `ShapeCastable`. This leads to misleading inferred types when developing components which are generic in arbitrary `ShapeLike`s - like current `MemoryBank`.)

> Maybe similar method could be used here?

Maybe. But please consider again if `ValueCastable`, `ShapeCastable` or something like this could be used here. It would be preferable to use standard Amaranth methods for customizing the language.

tilk:
Another thought: maybe handling everything at type level via `typing.Annotated` could work for this?

tilk (approved):

PR #47, opened 2025-01-30
Orphaned simultaneous transactions fix

### LINKED ISSUE #45: `branch` is always executed when the outer method is pruned
Consider the following code:
```
@def_method(m, self._resume_from_unsafe)
def _(pc):
    with condition(m, nonblocking=True) as branch:
        with branch(True):
            self.redirect_frontend(m, pc=pc)
```
and in the case where there are no callers of `_resume_from_unsafe`, `redirect_frontend` will be called every cycle. I presume this is because the outer method is pruned, but the underlying transaction is left untouched and is not guarded by `run` signal of the outer method anymore. 

The transaction profile confirms that the method `_resume_from_unsafe` is gone and the inner method is called every cycle:
```
$ scripts/tprof.py -m methods -g test/__profiles__/test.test_core.py\:\:TestCoreBasicAsm_2_fibonacci_mem_tiny\:\:test_asm_source_0.json  | egrep  "resume|redirect"
FetchUnit_redirect                                                              coreblocks/frontend/fetch/fetch.py:58                                        0    250
- StallController__resume_from_unsafe_cond0                                     coreblocks/frontend/stall_controller.py:92                                   0    250

```

tilk:
It turns out that, due to how simultaneous transactions are implemented, a transaction nested in and simultaneous with a method is run if the method has no callers. This PR detects such situations and disables the transaction. Maybe a little hacky, but it's the simplest solution I came up with.

Fixes #45.

### DIFF
--- a/transactron/core/manager.py
+++ b/transactron/core/manager.py
@@ -261,13 +261,21 @@ class TransactionManager(Elaboratable):
         simultaneous = set[frozenset[TBody]]()
 
         for elem in method_map.methods_and_transactions:
+            pruned_sim = False
             for sim_elem in elem.simultaneous_list:
+                if sim_elem not in method_map.methods_and_transactions:
+                    if sim_elem.independent_list:
+                        raise RuntimeError("Pruned method with independent list not supported")
+                    pruned_sim = True
+                    continue
                 for tr1, tr2 in product(method_map.transactions_for(elem), method_map.transactions_for(sim_elem)):
                     if tr1 in independents[tr2]:
                         raise RuntimeError(
                             f"Unsatisfiable simultaneity constraints for '{elem.name}' and '{sim_elem.name}'"
                         )
                     simultaneous.add(frozenset({tr1, tr2}))
+            if pruned_sim and elem in method_map.transactions:
+                elem.ready = Signal()
 
         # step 2: transitivity computation
         tr_simultaneous = set[frozenset[TBody]]()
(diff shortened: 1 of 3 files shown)


transactron/core/manager.py
```diff
@@ -261,13 +261,21 @@ def _simultaneous(self):
         simultaneous = set[frozenset[TBody]]()
 
         for elem in method_map.methods_and_transactions:
+            pruned_sim = False
             for sim_elem in elem.simultaneous_list:
+                if sim_elem not in method_map.methods_and_transactions:
```
lekcyjna123, line 266: Do I understand correctly, that if we have multiple elements simultaneous to each other, then every has to execute in the same cycle. So it will be enough so that one of them can not be called, so that whole set will be never called?

transactron/core/manager.py
```diff
@@ -261,13 +261,21 @@ def _simultaneous(self):
...
+                    if sim_elem.independent_list:
+                        raise RuntimeError("Pruned method with independent list not supported")
+                    pruned_sim = True
+                    continue
```
lekcyjna123, line 270: Why `continue` instead of `break`?

transactron/core/manager.py
tilk: Yes. It's not true when `independent_list` is used, which happens only in `Method.simultaneous_alternatives` (and, therefore, in `condition`). The condition below checks for that and raises an exception in this case, so that it's clear this case is currently not supported.

transactron/core/manager.py
tilk: To check for cases when the exception needs to be raised.

lekcyjna123 (approved):

transactron/core/manager.py
lekcyjna123: Thanks. I forgot about that case.

PR #48, opened 2025-01-31
Tprof as script in pyproject.toml

### LINKED ISSUE #31: Executables in Transactron
Transactron currently has an executable script which is useful in projects using it - `tprof.py`. It should be packaged as an executable script using `pyproject.toml` (see [packaging docs](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#creating-executable-scripts)).

tilk:
Fixes #31.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,9 @@ dev = [
     "pre-commit == 2.16.0"
 ]
 
+[project.scripts]
+tprof = "transactron.cmd.tprof:main"
+
 [tool.setuptools_scm]
 
 [tool.pyright]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -40,6 +40,9 @@ dev = [
     "pre-commit == 2.16.0"
 ]
 
+[project.scripts]
+tprof = "transactron.cmd.tprof:main"
```
lekcyjna123, line 44: I would suggest using `transactron-prof` as a command line name. It is longer but there will be lower probability of conflicts with other installed scripts.

lekcyjna123 (approved):

PR #53, opened 2025-02-04
One-hot mux circuit

tilk:
This PR adds a `OneHotMux` module, which is internally implemented using `OneHotSwitch`. This allows some cases where a one-hot mux is needed to be expressed cleaner.

Side note: it needs to be checked of `OneHotSwitch` actually reliably generates Yosys `$pmux` cells, as a lot has happened in Amaranth since this was last checked. I can't find any mention of reliable one-hot muxes in Amaranth codebase, so we should probably ask Amaranth devs about it.


tilk:
> it needs to be checked of `OneHotSwitch` actually reliably generates Yosys `$pmux` cells

I checked and unfortunately, it doesn't.

PR #57, opened 2025-02-08
Update Pyright

tilk:
Updated Pyright to 1.1.393, the most recent version as of today.

### DIFF
--- a/transactron/lib/reqres.py
+++ b/transactron/lib/reqres.py
@@ -1,4 +1,6 @@
 from amaranth import *
+
+from transactron.utils.transactron_helpers import from_method_layout
 from ..core import *
 from ..utils import SrcLoc, get_src_loc, MethodLayout
 from .connectors import Forwarder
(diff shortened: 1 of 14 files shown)


transactron/lib/reqres.py
```diff
@@ -63,8 +65,8 @@ def __init__(self, args_layout: MethodLayout, results_layout: MethodLayout, src_
             Alternatively, the source location to use instead of the default.
         """
         self.src_loc = get_src_loc(src_loc)
-        self.results_layout = results_layout
-        self.args_layout = args_layout
+        self.results_layout = from_method_layout(results_layout)
```
piotro888, line 68: method layout -> Layout in docstring

transactron/utils/amaranth_ext/shifter.py
```diff
@@ -287,7 +287,7 @@ def generic_shift_vec_left(
     Sequence[Value | ValueCastable]
         The shifted sequence, the same length as `data1`.
     """
-    return list(reversed(generic_shift_vec_right(list(reversed(data1)), list(reversed(data2)), offset)))
+    return list(reversed(generic_shift_vec_right(list(reversed(data1)), list(reversed(data2)), offset)))  # type: ignore
```
piotro888, line 290: Remove ValueCastable from data[1|2] method and in docstring, the same in shift_vec_right docstring

piotro888 (approved):
Looks ok, minor comments (github unfortunately doesn't let me add review on lines or suggest changes there)

transactron/lib/reqres.py
tilk: The `method layout` wording is used in other places, including the core. I think this is outside the scope of the PR.

PR #60, opened 2025-02-18
Min value

tilk:
This PR adds a function `min_value`, which finds the minimum value of some set of values. The algorithm used is inspired by I2C's arbitration protocol: bits are checked sequentially from MSB to LSB, and if some value has a 0 on a given position, then all values with 1 are excluded from further checks.

Based on #59.

### DIFF
--- a/transactron/utils/amaranth_ext/functions.py
+++ b/transactron/utils/amaranth_ext/functions.py
@@ -17,7 +17,7 @@ __all__ = [
     "flatten_signals",
     "shape_of",
     "const_of",
-    "min_value"
+    "min_value",
 ]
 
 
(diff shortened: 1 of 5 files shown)


transactron/utils/amaranth_ext/functions.py
```diff
@@ -132,3 +133,26 @@ def const_of(value: int, shape: ShapeLike) -> Any:
...
+    # extend inputs to constant width
+    new_values = list(Signal.like(result) for _ in values)
+    for sig, value in zip(new_values, values):
+        m.d.comb += sig.eq(value)
+    values = new_values
+
+    for i in reversed(range(0, len(result))):
+        res = Signal()
+        m.d.comb += res.eq(Cat(value[i] for value in values).all())
+        m.d.comb += result[i].eq(res)
```
Arusekk, line 152: I know this PR is from last year, but in case we want to adopt it, we might as well stick to idiomatic iteration:

```suggestion
    for i, res_i in reversed(enumerate(result)):
        res = Signal()
        m.d.comb += res.eq(Cat(value[i] for value in values).all())
        m.d.comb += res_i.eq(res)
```

PR #61, opened 2025-02-26
Superscalar metrics

tilk:
This PR modifies metrics to enable superscalar use - changing a metric multiple times in a single clock cycle. This is needed to be able to create superscalar circuits which utilize metrics.

The PR introduces a small syntactic change to method calls: the `enable` signal is renamed to `enable_call` and can be used as a keyword argument. This means that `enable_call` is now a reserved argument name. This was done to be able to remove method wrapper functions from metrics modules, which utilized an optional `cond` argument - which is basically `enable`. I'm not extra happy with the solution, but it gets the job done.

TODO:
* [x] Update metrics tests to test superscalar use.

### DIFF
--- a/transactron/lib/metrics.py
+++ b/transactron/lib/metrics.py
@@ -374,8 +374,6 @@ class TaggedCounter(Elaboratable, HwMetric):
         Transactron module
     tag: ValueLike
         The tag of the counter.
-    cond: ValueLike
-        When set to high, the counter will be increased. By default set to high.
     """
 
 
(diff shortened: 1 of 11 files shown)


transactron/lib/metrics.py
```diff
@@ -322,40 +344,37 @@ def elaborate(self, platform):
...
+        for tag_value, counter in self.counters.items():
+            m.d.sync += counter.value.eq(counter.value + popcount(runs[tag_value]))
 
-    def incr(self, m: TModule, tag: ValueLike, *, cond: ValueLike = C(1)):
-        """
-        Increases the counter of a given tag by 1.
+        return m
 
-        Should be called in the body of either a transaction or a method.
+    incr: Methods
```
lekcyjna123, line 365: I think that something went wrong here. Why this docstring is in such weird place?

lekcyjna123 (approved):

PR #77, opened 2025-06-24
Click I/O -> BasicIO

tilk:
This PR aims to make the ancient and largely unused `button click` modules more usable. This is done by adding configurability:

* Synchronization is made optional. This way, I/O classes can be used for controlling synchronous non-Transactron modules.
* Instead of rising edge only triggering, edge and level triggering with both polarities can be used. The possibility of usage with no triggering at all is also added.

The idea is that these modules could be used in the beginning of the Transactron tutorial.

TODO:
* [x] Documentation
* [x] Test


transactron/lib/basicio.py
```diff
@@ -0,0 +1,207 @@
...
+    If the `trigger` signal is not supplied, the default triggering logic
+    (low level) implies that the trigger is always active.
+
+    Inputs are optionally synchronized, which can be used for sampling
+    asynchronous signals from e.g. FPGA inputs.
+
+    Attributes
+    ----------
+    trigger: Signal, in
+        The button input.
```
lekcyjna123, line 148: Left over from buttons

transactron/lib/basicio.py
```diff
@@ -0,0 +1,207 @@
...
```
lekcyjna123, line 61: Left over from button.

transactron/lib/basicio.py
```diff
@@ -0,0 +1,207 @@
...
+        edge: bool = True,
+        polarity: bool = True,
+        synchronize: bool = False,
+        src_loc: int | SrcLoc = 0,
+    ):
+        """
+        Parameters
+        ----------
+        layout: method layout
+            The data format for the `dat` output and `put` method arguments.
```
lekcyjna123, line 180:
```suggestion
            The data format for the `data` output and `put` method arguments.
```

lekcyjna123 (approved):

PR #79, opened 2025-07-22
Stop accepting methods as constructor parameters

tilk:
This PR modifies library components which accept methods as parameters to do it instead in a separate static method `create`, constructors accept layouts instead. This way, `SimpleTestCircuit` can be used in tests, eliminating some boilerplate code. This will also allow implementing an abstraction similar to `lib.wiring` in the future.

As part of this change, the constructor and the `create` method of `Adapter` swapped meanings, but this was required for consistency.

Also, `ManyToOneConnectTrans` became `CrossbarConnectTrans`, as it bugged me for some time now that it was limited for no reason, and as the construction interface is changing anyway, I put the change here.


piotro888:
I'm not entirely sure about that, but maybe just because I'm used to old style. Any opinions from others?

At least about I think `proxy` could be renamed to something more obvious, if it will become a default way of passing methods.

tilk:
> I'm not entirely sure about that, but maybe just because I'm used to old style. 

I have some irrational resistance about this change too, and some rational (more LoC, but most of it come from the docstrings). But rationally, I believe that avoiding passing methods to constructors is a good change:

* In the presence of cycles, `proxy` allows conveniently breaking them.
* Even without cycles, refactoring can change dependencies between modules, forcing to reorder code.
* Compatible with `SimpleTestCircuit`.
* In cases when passing methods to constructors is very convenient (like some library code), the original behavior is still available by adding some boilerplate (the `create`  static methods).
* Connecting signals in Amaranth is done in a similar way (via `eq` or `lib.wiring`). Therefore, methods become more like interfaces.
* [Chisel actually does something similar](https://www.chisel-lang.org/docs/explanations/functional-module-creation).

> At least about I think `proxy` could be renamed to something more obvious, if it will become a default way of passing methods.

Yes, I actually planned to do that soon, my current best idea is `provide`.

piotro888 (approved):
Thanks for explanation, on a second thought it makes a lot of sense

piotro888:
I like the `provide` name too

PR #82, opened 2025-07-25
Refactor testing

### LINKED ISSUE #66: disussion: dev dependency group assignment
In my opinion `[dev]` should contain only dependencies needed for development of `transactron` itself.

Currently, when only installing required dependencies, full development of projects using `transactron` is not possible - unit test features from `transactron.testing` don't work.
We should move some of [dev] dependencies to main group or/and be more strict in imports from library code.

tilk:
This PR:

* Refactors `testing.infrastructure` into `testing.simulator` and `testing.test_case`.
* Moves development dependencies to dependency groups.
* Adds missing `__all__` declarations in `testing` submodules.
* Adds an `__all__` declaration to `testing`, limiting `from transactron.testing import *` to the most crucial definitions.
* Makes `TestCaseWithSimulator` a lazy import because of an optional dependency on `pytest`.

Fixes #66.


qbojj:
From what I understand latest python conventions, the development requirements (the packages that are only required for development, not library features), should be in `[dependency-groups]`. It prevents those packages from being listed as "extra" features on sites like PyPI.

Some tools also pull some dependency-groups by default into their venv (e.g `uv` pulls `dev` group)

See https://peps.python.org/pep-0735/

.github/workflows/main.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip3 install .[dev]
+          pip3 install --group dev .
```
qbojj: probably should be `pip3 install --group dev .[testing]'. Currently it doesn't really matter, but if some package that would not be in dev were to be added to `testing` this would fail

qbojj (approved):

qbojj (approved):

transactron/testing/simulator.py
```diff
@@ -0,0 +1,121 @@
...
+            await sim.delay(clk_period * max_cycles)
+            assert False, "simulation timed out"
+
+        self.add_testbench(timeout_testbench, background=True)
+
+    def run(self) -> None:
+        with self.ctx:
+            super().run()
+
+    def add_mock(self, val: MethodMock):
```
Arusekk, line 93: I like this move

transactron/testing/simulator.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+async def random_wait(ctx: SimulatorContext, max_cycle_cnt: int, *, min_cycle_cnt: int = 0):
+    """
+    Wait for a random amount of cycles in range [min_cycle_cnt, max_cycle_cnt]
+    """
+    await tick(ctx, random.randrange(min_cycle_cnt, max_cycle_cnt + 1))
+
+
+async def random_wait_geom(ctx: SimulatorContext, prob: float = 0.5, max_cycle_cnt: int = 2**16):
```
Arusekk, line 115: In the future, we might consider first rolling a number from Poisson distribution, and then waiting once. Of course only if it turns out performance-sensitive (for low probabilities).

pyproject.toml
Arusekk: We might as well use groups nesting:

```suggestion
lint = [
"pyright == 1.1.393",
"flake8 == 7.0.0",
"black == 24.4.2",
]
test = [
"pytest == 8.0.0",
"pytest-xdist == 3.5.0",
"hypothesis == 6.99.6",
]
docs = [
"pep8-naming == 0.13.3",
"Sphinx == 8.2.3",
"sphinx-rtd-theme == 3.0.2",
"myst-parser == 4.0.1",
"numpydoc == 1.5.0",
"sphinxcontrib-mermaid == 0.8.1",
]
dev = [
"pre-commit == 2.16.0",
{include-group = "lint"},
{include-group = "test"},
{include-group = "docs"},
```

We can then install the respective subsets in specific workflows, saving setup time in CI.

transactron/testing/__init__.py
```diff
@@ -1,11 +1,43 @@
...
+    "CallTrigger",
+    "SimpleTestCircuit",
+    "TestbenchIO",
+    "def_method_mock",
+    "data_const_to_dict",
+    "data_layout",
+]
+
+
+def __getattr__(name):
```
Arusekk, line 42: Looks like a hack, do we really need a lazy import? Or is pytest actually optional, so *testing* works without it (for example with stdlib `unittest`)?

Maybe we can prevent pytest on `from transactron import *` by just excluding `testing` from `transactron.__all__`?

transactron/testing/test_case.py
Arusekk: This is the only actual use of pytest vs testsuite-agnostic. Maybe we can drop it altogether, moving this init code to some other place?

transactron/testing/__init__.py
tilk: Some parts of `testing` don't require any unit testing framework at all, they just combine with the Amaranth simulator or code generator. I consider the `pytest` integration to be mostly convenience and reduce the reliance on it.

transactron/testing/simulator.py
tilk: Good idea, maybe make an issue out of it?

transactron/testing/test_case.py
tilk: Huh. I guess you're right. Maybe just good ol' subclassing, and have the pytest-extended version in a separate file? This way, most of the convenience is still usable without `pytest`.

pyproject.toml
tilk: Just a few seconds, but grouping dev tools is pretty cool actually.

transactron/testing/__init__.py
tilk: By the way, making a sensible `transactron.__all__` looks like a nice polishing touch thing to do.

transactron/testing/simulator.py
tilk: Actually, why Poisson distribution in particular?

PR #89, opened 2025-08-01
Rename request/grant to ready/run

tilk:
I believe that having distinct names for handshake signals for transactions (`request`/`grant`) and methods (`ready`/run`) is confusing and unneeded:

1. Many docstrings already in the code actually explain the transaction signals using method terminology.
2. In the current implementation, these signals are actually the same under the hood - both transactions and methods are implemented in terms of bodies, and sometimes a transaction body can actually become a method body (simultaneous transactions).
3. Transactions are in practice implemented rarely than methods (in Coreblocks, 60 transactions vs 150 methods), so it's easy to forget the names used in transactions.

Therefore I believe that it's time to change the user-facing names.

### DIFF
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -230,12 +230,12 @@ def _(foo: Value, bar: Value):
 
 ### Readiness signals
 
-If a transaction is not always ready for execution (for example, because of the dependence on some resource), a `request` parameter should be used.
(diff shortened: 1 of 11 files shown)


piotro888 (approved):
+1 for the change

PR #93, opened 2025-08-07
Rename proxy to provide

tilk:
As discussed in #79, `proxy` is now intended to be a mechanism for providing method implementations to components which call external methods. With its new important function, it's time to give it a more descriptive name.


piotro888 (approved):
Approved, as discussed in #79

PR #106, opened 2025-10-24
lib.connectors: add `peek` methods, unify clean->clear

awariac:
Adds `peek` for two stateful connectors:  `Forwarder` and `Pipe`.

It can be demonstrated as useful eg. here: https://github.com/kuznia-rdzeni/coreblocks/blob/67c2e0c84b3e6ae81d130ec3f22494f3378ff3d4/coreblocks/func_blocks/fu/lsu/dummyLsu.py#L102 where `Forwarder` read results are weirdly escaped from Transaction to other scopes via `av_comb`.

Extra for `Pipe` - renames `clean` to `clear` to be consistent with other modules.  Removed `head` as this is replaced by `peek`. Added `src_loc`

### DIFF
--- a/transactron/lib/connectors.py
+++ b/transactron/lib/connectors.py
@@ -97,6 +97,8 @@ class Forwarder(Elaboratable):
     ----------
     read: Method
         The read method. Accepts an empty argument, returns a structure.
+    peek: Method, nonexclusive
+        Like `read`, but doesn't take the value from `Forwarder`.
     write: Method
         The write method. Accepts a structure, returns empty result.
     """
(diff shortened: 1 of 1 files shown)


tilk:
> It can be demonstrated as useful eg. here: https://github.com/kuznia-rdzeni/coreblocks/blob/67c2e0c84b3e6ae81d130ec3f22494f3378ff3d4/coreblocks/func_blocks/fu/lsu/dummyLsu.py#L102 where `Forwarder` read results are weirdly escaped from Transaction to other scopes via `av_comb`.

Right, this is super weird, as the signals are escaped backwards, to the transaction's own `ready` signal. Some of the weirdness is currently unavoidable, as the ready signal does not come from a transaction context, but at least the ordering can be fixed.

Issue #105, opened 2025-10-14
Conditional calls in Simultaneous Transactions bug


awariac:
> This is indeed a Transactron problem. Simultaneous transactions work at the level of transactions, by joining them together into larger ones. When adding simultaneity constraints to methods, this actually has the same effect as adding simultaneity constraints to the transactions which call these methods. This is incorrect when the method call is conditional, and causes the simultaneous transaction to run even when the call is inactive.
>
> Adding the forwarder works around the issue, because then the simultaneous transaction constraint is created with the new `ConnectTrans`, which doesn't have conditional calls.
>
> There is no simple fix, some amount of redesigning the transaction manager will be needed to correct this problem. For now, I suggest the following course of action:
> 
> Try to add a check in Transactron to detect the problem early.

_Originally posted by @tilk in https://github.com/kuznia-rdzeni/coreblocks/issues/843#issuecomment-3350907439_

piotro888:
@tilk can you look into adding this error pls? I hit this bug for a second time in a different place today :/

PR #107, opened 2025-10-31
Detect simultaneous constraints with conditionally called methods

tilk:
Simultaneous constraints with conditionally called methods do not work correctly with the current implementation of simultaneous transactions (#105).  This PR adds a check which detects such situations and raises an exception, so that people don't waste time on weird issues caused by this problem. The restriction is planned to be (partially?) lifted when the issue is understood better.


piotro888 (approved):
makes sense for me, thanks again!

Issue #111, opened 2026-01-28
Interoperability with `amaranth.lib.stream`


qbojj:
`amaranth.lib.stream` provides a simple, standardized handshake to transfer data similar to single consumer methods.

Some modules could be added to increase interoperability with modules outside of transactron that use streaming interfaces.

This could be made in `ConnectTrans` style, or `Connect` style.

PR #112, opened 2026-02-03
add interoperability with Amaranth's streams

qbojj:
implements #111

### DIFF
--- a/test/lib/test_stream.py
+++ b/test/lib/test_stream.py
@@ -1,4 +1,5 @@
 import random
+from collections import deque
 
 from amaranth import *
 from amaranth.lib import stream, wiring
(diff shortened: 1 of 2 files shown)


test/lib/test_stream.py
piotro888: this looks like a lot of "manual" test cases.
While it may be important to test some fixed properties, we prefer fully randomized tests in the project (also they are more maintainable).

transactron/lib/stream.py
```diff
@@ -0,0 +1,192 @@
...
+from amaranth_types import ShapeLike, HasElaborate
+
+from ..core import *
+from ..utils import SrcLoc, get_src_loc
+from ..utils.data_repr import data_layout
+
+
+__all__ = [
+    "StreamSink",
+    "StreamSource",
```
piotro888: missing `StreamModuleWrapper`

transactron/lib/stream.py
```diff
@@ -0,0 +1,192 @@
...
+]
+
+
+class StreamSink(wiring.Component):
+    """Adapter from stream sink to method.
```
piotro888: please mention that this refers to Amaranth's lib.stream streams, also in `StreamSource`

piotro888 (commented):
Overall looks good, thanks for your contribution!
I left come comments.

transactron/lib/stream.py
```diff
@@ -0,0 +1,192 @@
...
+            How many stack frames deep the source location is taken from.
+            Alternatively, the source location to use instead of the default.
+        """
+        super().__init__(
+            {
+                "i": In(stream.Signature(shape)),
+            }
+        )
+
+        method_layout = data_layout(shape)
```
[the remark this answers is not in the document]
tilk: Indeed, it does look like this.

piotro888 (approved):
Thanks!

PR #115, opened 2026-02-22
Add Transactron talk video to docs

awariac:
a shameless plug

; more seriously, i think it may be a helpful introduction resource for some people.


Arusekk:
I think this can safely be put into README.

tilk:
Let's do it then.

PR #116, opened 2026-02-26
Add read_on_resp to MemoryBank

tilk:
This PR adds yet another parameter to `MemoryBank`, `read_on_resp`. It changes the read semantics of `MemoryBank` as follows:

* If false, original semantics is used: the value returned on response is the value present in the memory when the *request* method was called.
* If true, the value returned on response is the value present in the memory when *response* method was called.

To match this semantics, `transparent` with `read_on_resp` makes the response method return the value written to the same address in the same clock cycle, i.e. combinational forwarding.

Motivation: fix possible undetected RF bug in Coreblocks (https://github.com/kuznia-rdzeni/coreblocks/issues/831).

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ Transactron is inspired by [Bluespec](https://github.com/B-Lang-org/bsc) and its
 However, while Bluespec requires to follow its paradigm strictly down to simple register assignments, Transactron's abstractions are intended to be used at module boundaries, while the actual logic is written in plain Amaranth.
 Transactron is just a library and, as such, allows smooth interoperation with plain Amaranth HDL code.
 
+A short video introduction to the Transactron library and its concepts is available [here](https://video.infosec.exchange/w/kLKCRGDVQTegfdCqjiGQaK) ([ORConf 2025](https://fossi-foundation.org/orconf/2025)).
+
 ## State of the project
 
 The library is in alpha stage of development, but is already well tested because it serves a foundation for the [Coreblocks](https://github.com/kuznia-rdzeni/coreblocks) out-of-order RISC-V CPU.
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -19,6 +19,8 @@ In this way, access to methods is coordinated via the transaction system to avoi
 Methods can communicate with their callers in both directions: from caller to method and back.
 The communication is structured using Amaranth structures.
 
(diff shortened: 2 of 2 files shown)


transactron/lib/storage.py
```diff
@@ -100,20 +106,55 @@ def elaborate(self, platform) -> TModule:
...
+            if self.read_on_resp:
+                for j in range(self.writes_ports):
+                    m.d.comb += read_output_addr_match[i][j].eq(
+                        write_port[j].en & (write_port[j].addr == read_output_addr[i])
+                    )
+                    m.d.comb += overflow_addr_match[i][j].eq(
+                        write_port[j].en & (write_port[j].addr == overflow_addr[i])
+                    )
+                m.d.comb += read_output_value[i].eq(
+                    OneHotMux.create(
```
piotro888: Side note, I think `OneHotMux.create` should have named arguments (or at least the default value) for clarity.

transactron/lib/storage.py
```diff
@@ -100,20 +106,55 @@ def elaborate(self, platform) -> TModule:
...
-        # The first result is stored in the overflow buffer, the second - in the read value buffer of the memory.
-        # If the responses are always read as they arrive, overflow is never written and no stalls occur.
+        overflow_value = [Signal(self.shape) for _ in range(self.reads_ports)]
+        overflow_addr = [Signal(range(self.depth)) for _ in range(self.reads_ports)]
+        overflow_addr_match = [Signal(self.writes_ports) for _ in range(self.reads_ports)]
 
         for i in range(self.reads_ports):
+            if self.read_on_resp:
+                for j in range(self.writes_ports):
+                    m.d.comb += read_output_addr_match[i][j].eq(
```
piotro888: This is only used locally with argumet `[i]`.
Maybe it would be cleaner to create plain `Signal` for `*_match` under the `for`?
(can be subjective though)

transactron/lib/storage.py
```diff
@@ -100,20 +106,55 @@ def elaborate(self, platform) -> TModule:
...
+                        write_port[j].en & (write_port[j].addr == overflow_addr[i])
+                    )
+                m.d.comb += read_output_value[i].eq(
+                    OneHotMux.create(
+                        m,
+                        [(read_output_addr_match[i][j], write_port[j].data) for j in range(self.writes_ports)],
+                        read_port[i].data,
+                    )
+                )
+                m.d.comb += overflow_value[i].eq(
```
piotro888: I find the `value/data` naming a bit confusing. Any ideas for something better? In my code I sometimes used `_comb`, but I'm not a big fun of that too.

PR #126, opened 2026-03-11
implement initial pipeline builder

qbojj:
This module adds a pipeline builder - a pipelined design helper.

The pipeline builder creates live signals at each stage and passes them parallel to method invocations.

It allows to simplify passing the data parallel to some stages - e.g
-> instruction fetch - generates instruction info [imm, ISA regs, ...]
-> reg alloc -> consumes ISA regs and generates HW regs
-> instruction dispatch -> consumes HW regs, imm, ...

In this example the ISA regs would not be saved in the queues between reg alloc and instruction dispatch (if there are no further users).

It also allows to simply say `fifo_depth=n` to add fifo buffering in between stages


transactron/tracing.py
```diff
@@ -11,7 +11,6 @@
 from amaranth_types import SrcLoc, HasElaborate
 from . import core
 
-
```
tilk, line 14: Use `scripts/lint.sh format`.

test/lib/test_pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def elaborate(self, platform):
+        m = TModule()
+
+        p = PipelineBuilder()
+        p.add_external(self.write)
+
+        @p.stage(m, o=[("data", unsigned(8))])
+        def _(data):
+            return {"data": data + 1}
```
tilk, line 33: I don't see any test with a method call in `@p.stage`.

test/lib/test_pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+
+        p.add_external(self.read)
+        m.submodules += p.finalize()
```
tilk: The `p.finalize()` call can be easily forgotten. Using context managers would help with this, but at a syntactic cost of another indentation.

tilk: Why is `PipelineBuilder` creating a new `TModule` instead of adding to the already existing one?

test/lib/test_pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+        p = PipelineBuilder()
+        p.add_external(self.write)
+
+        @p.stage(m, o=[("data", unsigned(8))])
+        def _(data):
+            return {"data": data + 1}
+
+        # Call adder.compute using current pipeline fields; its output
+        # overwrites 'data' in the pipeline (same field name).
+        p.call_method(adder.compute)
```
tilk, line 256: I don't know if I like this interface. This is very short, but a lot is implicit here, and changing the definition of `compute` could lead to unpredictable results.

tilk (commented):
Some quick comments, more detailed review will follow later, when I have the time to do so.

test/lib/test_pipeline.py
tilk: Idea: add a module parameter to the pipeline constructor, some elaboratable is implicitly added as a submodule, finalization happens in the elaborate method.

qbojj: I wanted to add somewhat meaningful names for internal submodules of the pipeline. If I did that, and the user defines two pipelines in one TModule, then there could be some name-collisions.

tilk: Makes sense.

tilk:
After the discussion on superscalar pipelines, I believe this can be left for a later PR, and solved without needlessly complicating the core.

qbojj:
It looks like CI fails not in my code

test/lib/test_pipeline.py
```diff
@@ -0,0 +1,531 @@
...
+        m.submodules.pipeline = p = PipelineBuilder()
+        p.add_external(self.write)
+
+        @p.stage(m, o=[("data", unsigned(8))])
+        def _(data):
+            return {"data": data + 1}
+
+        @p.stage(m, o=[("data", unsigned(8))])
+        def _(data):
+            return {"data": data + 2}
```
tilk: Modify operation so that the stages do not commute.

test/lib/test_pipeline.py
```diff
@@ -0,0 +1,531 @@
...
+
+        p.add_external(self.read)
+
+        return m
+
+
+class TestFifoPipeline(TestCaseWithSimulator):
+    def test_pipeline(self):
```
tilk, line 197: Test the condition when the FIFO overflows.

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,408 @@
...
+# ---------------------------------------------------------------------------
+# PipelineBuilder
+# ---------------------------------------------------------------------------
+
+
+class PipelineBuilder(Elaboratable):
+    """Helper class for building transactional pipelines.
+
+    Each node in the pipeline can be a function stage, a provided-method node,
+    or a called-method node.  Call :meth:`finalize` after adding all nodes to
```
tilk: Finalize no longer exists.

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,408 @@
...
+    build the hardware.
+
+    Parameters
+    ----------
+    m : TModule
+        The module in which the pipeline is elaborated.
+
+    Examples
+    --------
+    See the module-level docstring for a complete example.
```
tilk: No module-level docstring?

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,408 @@
...
+        or interact with submodules in the middle of the pipeline (e.g. send
+        a request to a submodule and later receive the response back into the pipeline).
+
+        Parameters
+        ----------
+        method : Method
+            The ``Method`` whose body the pipeline will define.
+        ready : ValueLike
+            Additional combinational ready condition (default: always ready).
+
```
tilk, line 163: Add doc for missing parameters.

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,408 @@
...
+        This is a convenience wrapper around :meth:`add_external` that creates a new Method for you.
+
+        Parameters
+        ----------
+        i : MethodLayout
+            The input layout of the created Method.
+        o : MethodLayout
+            The output layout of the created Method.
+        ready : ValueLike
+            Additional combinational ready condition (default: always ready).
```
tilk: Add doc for missing parameters.

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,408 @@
...
+        pipeline signals as keyword arguments (or as a single ``arg`` struct) and
+        must return a ``dict`` mapping output field names to Amaranth values, or
+        ``None`` when it produces no new fields.
+
+        Parameters
+        ----------
+        o : MethodLayout
+            Layout of the fields *produced* (or overwritten) by this stage.
+        ready : ValueLike
+            Optional combinational ready signal for the stage's transaction.
```
tilk: Document other parameters.

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,408 @@
...
+        Returns
+        -------
+        Callable
+            A decorator that wraps the stage function, returning a
+            :class:`_FuncNode` that can be further modified by :meth:`fifo`.
+        """
+
+        src_loc = get_src_loc(src_loc)
+
+        def decorator(func: Callable) -> None:
```
tilk: Improve typing

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,408 @@
...
+            params = signature(func).parameters
+            i_layout_from_pipeline: dict[str, ShapeLike] = dict()
+            for p in params.values():
+                if p.name == "arg":
+                    raise TypeError("Pipeline stage function cannot have a parameter named 'arg'")
```
tilk: Does it make sense to allow the standard convention when `i` is not `None`?

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,408 @@
...
+
+            self.call_method(method, ready=ready, no_dependency=no_dependency, src_loc=src_loc)
+
+        return decorator
+
+    def fifo(self, depth: int, fifo_type=fifo.SyncFIFO, *, src_loc: int | SrcLoc = 0):
+        src_loc = get_src_loc(src_loc)
+        if self._next_forwarder is not None:
+            raise RuntimeError("Fifo was added twice for the same stage")
+        self._next_forwarder = lambda layout: FIFO(layout, depth, fifo_type, src_loc=src_loc)
```
tilk: Does it make sense to use `BasicFifo`?
Does it make sense to clear pipelines? (If not immediately, decide later)

tilk (changes requested):

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+
+class _PipelineNodeProtocol(Protocol):
+    def get_required_fields(self) -> StructLayout: ...
+
+    def get_generated_fields(self) -> StructLayout: ...
+
+    def finalize(self, m: TModule, method: Method) -> None: ...
+
+
+class _ForwarderLike(HasElaborate, Protocol):
```
lekcyjna123: This Protocol looks rather generic. Shouldn't be moved to some utils?

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+
+    @dataclass
+    class _NodeInfo:
+        node: _PipelineNodeProtocol
+        src_loc: SrcLoc
+        ready: ValueLike
+        no_dependency: bool
+        forwarder: Callable[[MethodLayout], _ForwarderLike]
+
+    def __init__(self, allow_unused: bool = False, allow_empty: bool = False):
```
lekcyjna123, line 113: Please add docstring with description of `allow_unused` and `allow_empty`

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+        o: MethodLayout = (),
+        *,
+        i: Optional[MethodLayout] = None,
+        ready: ValueLike = C(1),
+        no_dependency: bool = False,
+        src_loc: int | SrcLoc = 0,
+    ) -> Callable:
+        """Decorator that register a function as a pipeline
+
+        Decorator that registers a function-based pipeline stage.
```
lekcyjna123: So it creates a stage or whole pipeline? These two sentences are a little bit contradictory.

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+                    raise TypeError(
+                        f"Pipeline stage function has parameter {p.name} " f"that is not a live signal in the pipeline"
+                    )
+
+                i_layout_from_pipeline[p.name] = self._live_signal_shapes[p.name]
+
+            if i is not None:
+                i_layout = from_method_layout(i)
+                # check if the input layout matches the i_layout_from_pipeline
+                if i_layout.members != i_layout_from_pipeline:
```
lekcyjna123, line 335: Should it be by default equality? I would think that the default check should be if one is contain in the other and the equality should be enabled by some additional parameter (e.g. "strict"). What do you think?

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+# ---------------------------------------------------------------------------
+# PipelineBuilder
+# ---------------------------------------------------------------------------
+
+
+class PipelineBuilder(Elaboratable):
+    """Helper class for building transactional pipelines.
+
+    Each node in the pipeline can be a function stage, a provided-method node,
+    or a called-method node.
```
lekcyjna123, line 76: It would be nice to have a short example in doc string how to use that class.

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+                    raise TypeError(
+                        "Pipeline stage function has an input layout that does not match "
+                        "the live signal shapes in the pipeline"
+                    )
+            else:
+                i_layout = from_method_layout(i_layout_from_pipeline.items())
+
+            o_layout = from_method_layout(o)
+
+            method = Method(name=f"pipeline_stage_{len(self._nodes)}", i=i_layout, o=o_layout, src_loc=src_loc)
```
lekcyjna123: Would it make sense to allow to pass to PipelineBuilder an additional `name ` argument so that it can be used here in method name creation?

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+            def_method(m, method)(func)
+
+            self.call_method(method, ready=ready, no_dependency=no_dependency, src_loc=src_loc)
+
+        return decorator
+
+    def fifo(self, depth: int, fifo_type=fifo.SyncFIFO, *, src_loc: int | SrcLoc = 0):
```
lekcyjna123: Doc string is lacking.

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def _add_node(self, node: _PipelineNodeProtocol, ready: ValueLike, no_dependency: bool, src_loc: SrcLoc) -> None:
+        gen = node.get_generated_fields().members
+        req = node.get_required_fields().members
+
+        for k, v in req.items():
+            if k not in self._live_signal_shapes:
+                raise ValueError(f"Signal {k} from {src_loc} is required but not provided")
+
+            if self._live_signal_shapes[k] != v:
```
lekcyjna123, line 553: Should we require here equality or `assign` compatibility would be enough?

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+                raise ValueError(
+                    f"Signal {k} from {src_loc} has incompatible shape: expected {v}, got {self._live_signal_shapes[k]}"
+                )
+
+        if no_dependency and req:
```
lekcyjna123: Shouldn't be this check before the loop?

transactron/lib/pipeline.py
```diff
@@ -0,0 +1,398 @@
...
+    def elaborate(self, platform) -> TModule:
+        m = TModule()
+
+        live_types = self.get_live_signals()
+
+        read_methods: list[Optional[Method]] = [None] * len(self._nodes)
+        write_methods: list[Optional[Method]] = [None] * len(self._nodes)
+
+        for i in range(len(self._nodes) - 1):
+            read_methods[i + 1] = Method(name=f"{i}_read", o=live_types[i].items())
```
lekcyjna123: Maybe?
```suggestion
            read_methods[i + 1] = Method(name=f"{i}_pipeline_read", o=live_types[i].items())
```

transactron/lib/pipeline.py
qbojj: `i_layout_from_pipeline` is already filtered by consumed fields. This check essentially checks if the shapes match

transactron/lib/pipeline.py
qbojj: I was thinking about it, but allowing only assign compatibility would probably introduce silent truncations. Could be a `no_conv` pipeline option though

transactron/lib/pipeline.py
qbojj: I don't really know where should I put this

tilk: `lib/connectors.py` seems to be good choice. Maybe define two protocols: `Connector` and `ClearableConnector`.

tilk (approved):

lekcyjna123 (approved):

PR #127, opened 2026-03-19
Better error message for layout mismatch

tilk:
This helps debug things when refactoring.

### DIFF
--- a/transactron/core/method.py
+++ b/transactron/core/method.py
@@ -146,7 +146,7 @@ class Method(TransactionBase["Transaction | Method"]):
             raise ValueError(
                 textwrap.dedent(
                     f"""\
-                    Method {value.name} defined at {value.src_loc} has different interface than {self.name}
+                    Method {value.name} at {value.src_loc} has different interface than {self.name} at {self.src_loc}
                     {value.name} input: {value.data_in.shape()}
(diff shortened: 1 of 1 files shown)


awariac:
can we use src loc too? Thats helpful for for methods like `read`

tilk:
Why not.

transactron/core/method.py
```diff
@@ -142,7 +143,17 @@ def _set_impl(self, value: "Body | Method"):
         if self._body_ptr is not None:
             raise RuntimeError(f"Method '{self.name}' already defined")
         if value.data_in.shape() != self.layout_in or value.data_out.shape() != self.layout_out:
-            raise ValueError(f"Method {value.name} has different interface than {self.name}")
+            raise ValueError(
+                textwrap.dedent(
+                    f"""\
+                    Method {value.name} defined at {value.src_loc} has different interface than {self.name}
```
awariac: what about `self.name` location?

tilk: For a moment I thought that's not very useful, but that's not really true - `_set_impl` is called from `provide`, and calling it on methods which are not locally defined is actually the typical use case. Corrected.

awariac (approved):

PR #128, opened 2026-04-01
add traceback when calling a single method in the same transaction

qbojj:
This PR adds traceback when a method is called two times from the same transaction. It detects cycles and two call trees with a common ancestor.

Was tested on test/core/test_methods.py

### DIFF
--- a/transactron/core/manager.py
+++ b/transactron/core/manager.py
@@ -1,6 +1,6 @@
 from collections import defaultdict, deque
 from collections.abc import Callable, Iterable, Sequence, Collection, Mapping
-from typing import Optional, TypeAlias
+from typing import TypeAlias
 from os import environ
 from amaranth import *
 from itertools import chain, filterfalse, product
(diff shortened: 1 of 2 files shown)


transactron/core/manager.py
```diff
@@ -36,20 +36,94 @@ def __init__(self, transactions: Iterable[Transaction]):
...
                 self.ancestors_by_call[(transaction, method)] = new_ancestors = (method, *ancestors)
                 rec(transaction, method, new_ancestors)
 
         for transaction in transactions:
-            self.methods_by_transaction[TBody(transaction._body)] = []
+            methods_by_transaction_internal[TBody(transaction._body)] = dict()
             rec(TBody(transaction._body), transaction._body, ())
+            self.methods_by_transaction = {
+                transaction: list(methods.keys()) for transaction, methods in methods_by_transaction_internal.items()
+            }
```
tilk: Shouldn't this be outside the loop?

transactron/core/manager.py
```diff
@@ -36,20 +36,94 @@ def __init__(self, transactions: Iterable[Transaction]):
...
+            if cycle is not None:
+                cycle = cycle + [method]
+                cycle_str = " -> ".join(f"{m.name} ({m.src_loc})" for m in reversed(cycle))
+                msg += f"\nCycle: {cycle_str}"
+            else:
+                path1, path2 = find_call_paths_from_tree(transaction, source, method)
+                path1 = [method] + path1
+
+                path1_str = " -> ".join(f"{m.name} ({m.src_loc})" for m in reversed(path1))
+                path2_str = " -> ".join(f"{m.name} ({m.src_loc})" for m in reversed(path2))
```
tilk: I believe `src_loc` is a tuple, so this should result with double parentheses.

transactron/core/manager.py
```diff
@@ -36,20 +36,94 @@ def __init__(self, transactions: Iterable[Transaction]):
...
+                path1.append(m1)
+                path2.append(m2)
+                m1 = methods_by_transaction_internal[transaction][MBody(m1)]
+                m2 = methods_by_transaction_internal[transaction][MBody(m2)]
+
+            # append the common ancestor
+            path1.append(m1)
+            path2.append(m2)
+
+            return path1, path2
```
tilk: I'm trying to understand what are you trying to achieve here. Isn't the information you need for error reporting already in `ancestors_by_call`?

tilk (approved):

awariac (approved):
nice improvement in error reporting :)

PR #133, opened 2026-04-06
fix nested transaction readiness

### LINKED ISSUE #132: Method call in nested transaction happens unconditionally
When a method call is used in a nested Transaction, that inner method call can happen without either of the transaction being run.

```python
from amaranth import *

from transactron import TModule, Transaction, Method, def_method

class Outer(Elaboratable):
    def elaborate(self, _):
        m = TModule()

        inner = Method()

        outer_trans_run = Signal(1)

        @def_method(m, inner)
        def _():
            m.d.sync += Print("Inner method")

        trans_ready = Signal(1)

        outer = Transaction()
        with outer.body(m, ready=C(0)):
            m.d.sync += Print("Outer transaction")
            outer_trans = Transaction()
            with outer_trans.body(m, ready=trans_ready):
                m.d.sync += Print("Inner transaction")
                inner(m)
            m.d.top_comb += outer_trans_run.eq(outer_trans.run)

        # Method definition shows the same issue
        # outer = Method()
        # @def_method(m, outer, ready=C(0))
        # def _():
        #     m.d.sync += Print("Outer method")
        #     outer_trans = Transaction()
        #     with outer_trans.body(m, ready=trans_ready):
        #         m.d.sync += Print("Transaction")
        #         inner(m)
        #     m.d.top_comb += outer_trans_run.eq(outer_trans.run)

        # m.d.comb += trans_ready.eq(outer.run) # This fixes the issue
        m.d.comb += trans_ready.eq(1)

        m.d.sync += Print(inner.run, outer_trans_run, outer.run)

        return m

if __name__ == '__main__':
    from amaranth.sim import Simulator
    from transactron import TransactronContextElaboratable
    p = Outer()
    m = TransactronContextElaboratable(p)
    sim = Simulator(m)
    sim.add_clock(1e-6)
    async def test(ctx):
        for _ in range(10):
            await ctx.tick()
    sim.add_testbench(test)
    sim.run()
```

Actual output:

```python
Inner method
1 0 0
Inner method
1 0 0
Inner method
1 0 0
Inner method
1 0 0
Inner method
1 0 0
Inner method
1 0 0
Inner method
1 0 0
Inner method
1 0 0
Inner method
1 0 0
Inner method
1 0 0
```

Same behavior when the out most level Transaction is an method definition (doesn't help if the method was not used or disabled).
Disabling the inner Transaction fixes the issue. Seems like the ready flag did not propagate down properly.

qbojj:
Closes #132

### DIFF
--- a/transactron/core/manager.py
+++ b/transactron/core/manager.py
@@ -1,5 +1,6 @@
 from collections import defaultdict, deque
 from collections.abc import Callable, Iterable, Sequence, Collection, Mapping
+import textwrap
 from typing import TypeAlias
 from os import environ
 from amaranth import *
@@ -111,6 +112,15 @@ class TransactionManager(Elaboratable):
         self.cc_scheduler = cc_scheduler
 
     @staticmethod
+    def _relations(method_map: MethodMap) -> Sequence[Relation]:
+        return [
+            Relation(start=elem, **dataclass_asdict(relation))
+            for elem in method_map.methods_and_transactions
+            for relation in elem.relations
+            if relation.end in method_map.methods_and_transactions  # prune relations with uncalled methods
+        ]
+
+    @staticmethod
     def _transactions_exclusive(method_map: MethodMap, trans1: TBody, trans2: TBody):
         tms1 = [trans1] + method_map.methods_by_transaction[trans1]
         tms2 = [trans2] + method_map.methods_by_transaction[trans2]
@@ -187,12 +197,7 @@ class TransactionManager(Elaboratable):
                     ):
                         add_edge(transaction1, transaction2, Priority.UNDEFINED, True)
 
-        relations = [
-            Relation(start=elem, **dataclass_asdict(relation))
-            for elem in method_map.methods_and_transactions
-            for relation in elem.relations
-            if relation.end in method_map.methods_and_transactions  # prune relations with uncalled methods
-        ]
+        relations = TransactionManager._relations(method_map)
 
         for relation in relations:
             start = relation.start
@@ -237,15 +242,10 @@ class TransactionManager(Elaboratable):
         return method_enables
 
     @staticmethod
-    def _ready_dependencies(method_map: MethodMap) -> Mapping[TBody, Sequence[ValueLike]]:
-        ready_dependencies = defaultdict[TBody, list[ValueLike]](list)
+    def _ready_dependencies(method_map: MethodMap) -> TransactionGraph:
+        ready_dependencies = defaultdict[TBody, set[TBody]](set)
 
-        relations = [
-            Relation(start=elem, **dataclass_asdict(relation))
-            for elem in method_map.methods_and_transactions
-            for relation in elem.relations
-            if relation.end in method_map.methods_and_transactions  # prune relations with uncalled methods
-        ]
+        relations = TransactionManager._relations(method_map)
 
         for relation in relations:
             if not relation.ready_dependent:
@@ -256,7 +256,7 @@ class TransactionManager(Elaboratable):
                     if trans_start is not trans_end and not TransactionManager._transactions_exclusive(
                         method_map, trans_start, trans_end
                     ):
-                        ready_dependencies[trans_end].append(trans_start.ready)
+                        ready_dependencies[trans_end].add(trans_start)
 
         return ready_dependencies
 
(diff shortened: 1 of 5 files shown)


transactron/core/transaction.py
```diff
@@ -127,7 +127,7 @@ def body(self, m: TModule, *, ready: ValueLike = C(1)) -> Iterator["Transaction"
         )
         self._set_impl(m, impl)
 
-        m.d.av_comb += impl.ready.eq(ready)
+        m.d.comb += impl.ready.eq(ready)
```
tilk: Similar thing is in `Method`, did you look at that?

Did this change impact Coreblocks in any way?

Why did the case without a method call work?

qbojj: It seems like the methods have the same problems:
```py
m = TModule()

outer = Method()
middle = Method()
inner = Method()

@def_method(m, inner)
def _():
    m.d.comb += self.t2.eq(1)

@def_method(m, outer, ready=self.r1)
def _():
    m.d.comb += self.t1.eq(1)

    @def_method(m, middle, ready=self.r2)
    def _():
        inner(m)

with Transaction().body(m):
    outer(m)

with Transaction().body(m):
    middle(m)

return m
```

qbojj:
I feel like this is a bad way to do this. I will explore other solutions

qbojj:
If this will not work I will be mad

qbojj:
All tests passed in both transactron and coreblocks

tilk:
My feeling is that this is a hack. For me, the mystery to solve is: in the original bug report, why is `inner.run` set while `outer_trans.run` is not? This should not happen, regardless of nesting.

transactron/core/transaction.py
qbojj: > Why did the case without a method call work?

In fact the problem was that the enable signal of the call itself was avoiding, and it should not have been. As such only calling inside of disabled transaction (by higher level transaction not running) was incorrectly signalled as running.

I still don't know exactly where the running signals are masked by the topmost transaction though.

qbojj:
> My feeling is that this is a hack. For me, the mystery to solve is: in the original bug report, why is `inner.run` set while `outer_trans.run` is not? This should not happen, regardless of nesting.

I don't know if this is in response to the previous fix, but I have changed the implementation

tilk:
> > My feeling is that this is a hack. For me, the mystery to solve is: in the original bug report, why is `inner.run` set while `outer_trans.run` is not? This should not happen, regardless of nesting.
> 
> I don't know if this is in response to the previous fix, but I have changed the implementation

This is still a hack. Enabled method call should not generate a run signal if the calling transaction's run signal is not set.

qbojj:
currently only the internal `Body` has the `run` signal generating - the user facing facilities are not. When I tried changing the `run` signal in the `Body` the lib/simultaneous got broken. So yeah, feels like a hack

yuyichao:
I'm sure you already know this but going through the closed bug report, the symptom seems to be extremely similar to https://github.com/kuznia-rdzeni/transactron/issues/45, though of course not necessarily the same underlying cause.

tilk:
This should probably be handled by an attribute on the conflict relation indicating a dependency on `ready` signals should be added by the manager. Such a relation will be automatically erased in simultaneous transaction elaboration.

Your approach is a stop-gap because, while it stops bad things from happening, it still allows the manager to run a transaction which will not really run because of nesting, possibly stealing cycles from other transactions which could run instead.

transactron/core/manager.py
```diff
@@ -381,6 +408,7 @@ def elaborate(self, platform):
 
             method_map = MethodMap(self.transactions)
             cgr, porder = TransactionManager._conflict_graph(method_map)
+            ready_dependencies = TransactionManager._ready_dependencies(method_map)
```
tilk: Does not throw exceptions, there is no need for silencing.

transactron/core/manager.py
```diff
@@ -233,6 +236,30 @@ def rec(transaction: TBody, source: Body):
...
+    @staticmethod
+    def _ready_dependencies(method_map: MethodMap) -> Mapping[TBody, Sequence[ValueLike]]:
+        ready_dependencies = defaultdict[TBody, list[ValueLike]](list)
+
+        relations = [
+            Relation(start=elem, **dataclass_asdict(relation))
+            for elem in method_map.methods_and_transactions
+            for relation in elem.relations
+            if relation.end in method_map.methods_and_transactions  # prune relations with uncalled methods
+        ]
```
tilk: This is copy-pasted, I see potential for bugs.

transactron/core/manager.py
```diff
@@ -233,6 +236,30 @@ def rec(transaction: TBody, source: Body):
 
         return method_enables
 
+    @staticmethod
+    def _ready_dependencies(method_map: MethodMap) -> Mapping[TBody, Sequence[ValueLike]]:
```
tilk: Maybe this function should construct a graph, instead of grabbing the signals directly?

transactron/core/manager.py
tilk: Actually, there is a thing to check: if a transaction is ready-dependent on another transactions which it is conflicting with, it will never run, which is certainly a bug and worth reporting. If `_ready_dependencies` returned a graph, this check could be done here, instead of `_ready_dependencies`.

tilk:
@yuyichao Could you verify this works for you?

yuyichao:
Yes, it seems to work for the cases I've tested. Thanks!

tilk (approved):

PR #138, opened 2026-04-09
Use _guardedcontextmanager in TModule

### LINKED ISSUE #136: Replicate amaranth Module's builtin protection
I spent a few hours debugging an logical error since I accidentally typed `if m.If` instead of `with m.If`.

The original amaranth `Module` appears to [have support to catch such errors](https://github.com/amaranth-lang/amaranth/blob/a30350dcab953f9aeb814c6ff721c634886e0691/amaranth/hdl/_dsl.py#L199), (added in https://github.com/amaranth-lang/amaranth/commit/9964fc6b5768654bfc6747e62ed351116d858a24)

```julia
In [4]: bool(Module().If(1))
---------------------------------------------------------------------------
SyntaxError                               Traceback (most recent call last)
Cell In[4], line 1
----> 1 bool(Module().If(1))

File /usr/lib/python3.14/site-packages/amaranth/hdl/_dsl.py:206, in _GuardedContextManager.__bool__(self)
    205 def __bool__(self):
--> 206     raise SyntaxError("`if m.{kw}(...):` does not work; use `with m.{kw}(...)`"
    207                       .format(kw=self.keyword))

SyntaxError: `if m.If(...):` does not work; use `with m.If(...)`
```

OTOH, the `TModule` version does not appear to have this,

```julia
In [5]: bool(TModule().If(1))
Out[5]: True
```

It'll be nice to have the same feature on `TModule`

tilk:
Uses an Amaranth internal, but so what, they are unlikely to change this, and in case they do fixing is easy.

Fixes #136.

### DIFF
--- a/transactron/core/tmodule.py
+++ b/transactron/core/tmodule.py
@@ -4,7 +4,7 @@ from amaranth import *
 from amaranth_types import StatementLike, ModuleLike, ValueLike, SwitchKey
 from typing import Optional, Self, NoReturn
 from contextlib import contextmanager
-from amaranth.hdl._dsl import FSM
+from amaranth.hdl._dsl import FSM, _guardedcontextmanager
 
 __all__ = ["TModule"]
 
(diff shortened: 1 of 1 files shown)


transactron/core/tmodule.py
Arusekk, line 161: Do we want to apply it here, too? Amaranth's implementation does not look bloated, so it should not hurt much. However, we only use it internally.

tilk: Indeed, this is internal, and I believe using `if` by mistake here has low probability. This PR is basically to match user expectations for people coming from Amaranth.

qbojj (approved):

PR #140, opened 2026-04-15
Switch to Python 3.13

tilk:
I'm doing all my work on Coreblocks and Transactron on Python 3.13 for some time with no issues. It's probably time to change it to be the default. Why not 3.14? Not yet available in stable Debian ;)

New generic type and type alias syntax can finally be used.

### DIFF
--- a/.github/workflows/deploy_gh_pages.yml
+++ b/.github/workflows/deploy_gh_pages.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.13"
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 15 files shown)


Arusekk:
I just checked that all of this syntax works fine in 3.12 as well, and the tests pass.
We do not have to explicitly support 3.12 though, especially if even Debian stable has 3.13.

Is there someone stuck with 3.12? Or are there any projects depending on transactron already?

tilk:
The only external user of Transactron I am aware of is @yuyichao. Are you okay with this change?

yuyichao:
Yes. I’m using 3.14 anyway….

lekcyjna123 (approved):

qbojj (approved):

tilk:
Want to first merge #146 and make a bugfix release.

PR #141, opened 2026-04-15
Allow for overlapping called method in semantically exclusive

### LINKED ISSUE #10: Exclusive branches in diamond call graphs
Suppose we have a code like this:
```python
@def_method(m, method1):
def _():
    method3(m)

@def_method(m, method2):
def _():
    method3(m)

with Transaction().body(m):
    with m.If(foo):
        method1(m)
    with m.Else():
        method2(m)
```
Currently, an exception "Method 'method3' can't be called twice from the same transaction" will be raised. But these calls are on syntactically different branches, so they could technically be allowed.

qbojj:
Closes #10

### DIFF
--- a/test/core/test_branches.py
+++ b/test/core/test_branches.py
@@ -194,6 +194,64 @@ class TestExclusiveDiamondCall(TestCaseWithSimulator):
             sim.add_testbench(run)
 
 
+class ExclusiveDiamondValidateArgumentsCircuit(Elaboratable):
+    def __init__(self):
+        self.sel = Signal()
+        self.method_left = Method()
+        self.method_right = Method()
+        self.method_inner = Method(i=[("value", 2)])
+        self.running = Signal()
+
+    def elaborate(self, platform):
+        m = TModule()
+
+        @def_method(m, self.method_inner, validate_arguments=lambda value: value == 0b01)
+        def _(value):
+            pass
+
+        @def_method(m, self.method_left)
+        def _():
+            self.method_inner(m, value=0b01)
+
+        @def_method(m, self.method_right)
+        def _():
+            self.method_inner(m, value=0b10)
+
+        with Transaction().body(m):
+            with m.If(self.sel):
+                self.method_left(m)
+            with m.Else():
+                self.method_right(m)
+
+            m.d.comb += self.running.eq(1)
+
+        return m
+
+
+class TestExclusiveDiamondValidateArguments(TestCaseWithSimulator):
+    def test_exclusive_diamond_validate_arguments(self):
+        dut = ExclusiveDiamondValidateArgumentsCircuit()
+        circ = SimpleTestCircuit(dut)
+
+        async def run(sim):
+            sim.set(dut.sel, 1)
+            await sim.tick()
+            assert sim.get(dut.running)
+            assert sim.get(dut.method_left.run)
+            assert not sim.get(dut.method_right.run)
+            assert sim.get(dut.method_inner.run)
+
+            sim.set(dut.sel, 0)
+            await sim.tick()
+            assert not sim.get(dut.running)
+            assert not sim.get(dut.method_left.run)
+            assert not sim.get(dut.method_right.run)
+            assert not sim.get(dut.method_inner.run)
+
+        with self.run_simulation(circ) as sim:
+            sim.add_testbench(run)
+
+
 class TestNonExclusiveDiamondCall(TestCaseWithSimulator):
     def test_nonexclusive_diamond_call(self):
         circ = SimpleTestCircuit(NonExclusiveDiamondCallCircuit())
--- a/test/core/test_branches.py
+++ b/test/core/test_branches.py
@@ -197,6 +197,7 @@ class TestExclusiveDiamondCall(TestCaseWithSimulator):
 class ExclusiveDiamondValidateArgumentsCircuit(Elaboratable):
     def __init__(self):
         self.sel = Signal()
+        self.filter_value = Signal(2)
         self.method_left = Method()
         self.method_right = Method()
         self.method_inner = Method(i=[("value", 2)])
@@ -205,7 +206,7 @@ class ExclusiveDiamondValidateArgumentsCircuit(Elaboratable):
     def elaborate(self, platform):
         m = TModule()
 
-        @def_method(m, self.method_inner, validate_arguments=lambda value: value == 0b01)
+        @def_method(m, self.method_inner, validate_arguments=lambda value: value == self.filter_value)
         def _(value):
             pass
 
@@ -234,19 +235,22 @@ class TestExclusiveDiamondValidateArguments(TestCaseWithSimulator):
         circ = SimpleTestCircuit(dut)
 
         async def run(sim):
-            sim.set(dut.sel, 1)
-            await sim.tick()
-            assert sim.get(dut.running)
-            assert sim.get(dut.method_left.run)
-            assert not sim.get(dut.method_right.run)
-            assert sim.get(dut.method_inner.run)
+            for filter_left in (False, True):
+                sim.set(dut.filter_value, 0b01 if filter_left else 0b10)
 
-            sim.set(dut.sel, 0)
-            await sim.tick()
-            assert not sim.get(dut.running)
-            assert not sim.get(dut.method_left.run)
-            assert not sim.get(dut.method_right.run)
-            assert not sim.get(dut.method_inner.run)
+                sim.set(dut.sel, 1)
+                await sim.tick()
+                assert sim.get(dut.running) == filter_left
+                assert sim.get(dut.method_left.run) == filter_left
+                assert not sim.get(dut.method_right.run)
+                assert sim.get(dut.method_inner.run) == filter_left
+
+                sim.set(dut.sel, 0)
+                await sim.tick()
+                assert sim.get(dut.running) != filter_left
+                assert sim.get(dut.method_left.run)
+                assert sim.get(dut.method_right.run) != filter_left
+                assert sim.get(dut.method_inner.run) != filter_left
 
         with self.run_simulation(circ) as sim:
             sim.add_testbench(run)
--- a/test/core/test_branches.py
+++ b/test/core/test_branches.py
@@ -1,3 +1,5 @@
+import pytest
+
 from amaranth import *
 from itertools import product
 from transactron.core import (
@@ -11,7 +13,7 @@ from transactron.core import (
 from transactron.core.tmodule import CtrlPath
 from transactron.core.manager import MethodMap
 from unittest import TestCase
-from transactron.testing import TestCaseWithSimulator
+from transactron.testing import SimpleTestCircuit, TestCaseWithSimulator
 from transactron.utils.dependencies import DependencyContext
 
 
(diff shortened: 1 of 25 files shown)


transactron/core/manager.py
```diff
@@ -28,22 +28,39 @@
 TransactionScheduler: TypeAlias = Callable[["MethodMap", TransactionGraph, TransactionGraphCC, PriorityOrder], Module]
 
 
+def call_paths_exclusive(path1: tuple[CtrlPath, ...], path2: tuple[CtrlPath, ...]):
```
tilk, line 33: Did you see `longest_common_prefix`? It's even used in this file.

transactron/core/manager.py
```diff
@@ -28,22 +28,39 @@
...
+        def report_bad_case(
+            transaction: TBody,
+            method: MBody,
+            first_ancestors: tuple[MBody, ...],
+            second_ancestors: tuple[MBody, ...],
+        ):
             msg = (
-                f"Method '{method.name}' {method.src_loc} called twice from "
-                + f"transaction '{transaction.name}' {transaction.src_loc}"
+                f"Method '{method.name}' {method.src_loc} called twice from transaction "
```
tilk: Could you please STOP IT with unexpected styling changes? Thank you.

test/core/test_branches.py
```diff
@@ -97,3 +99,87 @@ def test_conflict_removal(self):
 
         for s in cgr.values():
             assert not s
+
+
+class ExclusiveDiamondCallCircuit(Elaboratable):
```
tilk, line 105: Could you also verify that arguments are passed correctly in such a case?

tilk:
Please verify that this change works correctly with Coreblocks after removing the `condition` workarounds.

transactron/core/manager.py
qbojj: I had a problem with `CtrlPath` being unhashable

qbojj:
Branch without workarounds:
https://github.com/qbojj/coreblocks/tree/feat-s-mode-csrs-overlapping-methods

qbojj:
passes all tests: https://github.com/qbojj/coreblocks/pull/1

transactron/core/manager.py
tilk: By the way, it's pretty trivial to make `CtrlPath` hashable. But the change in `longest_common_prefix` is good - not only it removes the hashability requirement, it also should make it slightly more efficient.

transactron/core/manager.py
```diff
@@ -59,29 +72,45 @@ def report_bad_case(transaction: TBody, method: MBody, second_ancestors: tuple[M
...
+                    for i, old_path in enumerate(old_paths):
+                        if not call_paths_exclusive(old_path, new_call_path):
+                            report_bad_case(transaction, method, old_ancestors[i], new_ancestors)
+
+                    self.ancestors_by_call[(transaction, method)].append(new_ancestors)
+                    self.call_paths_by_call[(transaction, method)].append(new_call_path)
+                    if method not in self.methods_by_transaction[transaction]:
+                        self.methods_by_transaction[transaction].append(method)
+                        self.transactions_by_method[method].append(transaction)
+                        self.argument_by_call[(transaction, method)] = arg_rec
```
tilk: You only put the argument of the first call into `argument_by_call`. The only use of `argument_by_call` attribute is for `validate_arguments` - this feature probably is broken by this change. Please:

* Write a test which exhibits the breakage.
* Fix it.

tilk (changes requested):
Found a possible bug.

test/core/test_branches.py
```diff
@@ -97,3 +99,163 @@ def test_conflict_removal(self):
...
+    def elaborate(self, platform):
+        m = TModule()
+
+        @def_method(m, self.method_inner, validate_arguments=lambda value: value == 0b01)
+        def _(value):
+            pass
+
+        @def_method(m, self.method_left)
+        def _():
+            self.method_inner(m, value=0b01)
```
tilk, line 217: Test only validates one permutation, it could miss a bug.

qbojj:
The question is what should be the semantics of validate_arguments, as it is defined now the method call is always checked for valid arguments, so even if the method is not actually called, it needs valid arguments for the transaction to run.

What that would imply with multiple calls to the same method in the same transaction? - All call paths need to provide valid arguments for the transaction to run.

Now, do we want to preserve the semantics, or just do argument value checking on the executed paths?

IMO we should require all call paths to provide valid arguments (this would make sense, as all underlaying methods would normally need to be valid to be called anyway, and if the user wants otherwise, they should use `transactron.lib.simultaneous`)

tilk:
The intention of the `validate_arguments` feature is that a method can "reject" some calls. This way, it can "filter" them, so - if multiple calls are possible - a valid one can be selected among them. The feature is currently used in the retirement stage of Coreblocks, where the `precommit` method only allows calls for instructions which are about to be committed.

I think that the semantics you propose is the more surprising one: it makes a call which looks perfectly fine to be blocked because another call on an unselected path has a wrong argument. The argument from a disabled call is ignored, checking it seems weird and unexpected. Unfortunately, there is a possible interaction with transaction scheduling: validation is logically "before" scheduling (as the scheduler can select a transaction to run based on that info), while call selection is "after" scheduling (as disabled calls doesn't influence scheduling). It might be a necessary evil to perform the verification for disabled calls.

@lekcyjna123 do you have any thoughts?

qbojj:
I have read through the `validate_arguments` feature and what I found:
- It does actually validate the taken path.
- It only works well when method is called totally - the method call is called on every possible path. If that is not the case the `0` vector is checked instead if the method is not called in active path.

First of all I would like to unify the inside-single-body multiple calls and cross-body multiple calls.

I could do one of three things:
- Validate all call paths
- Validate only active path
- Validate only active path and if none present - validate with 0 (current behaviour)

tilk:
> It does actually validate the taken path.

Yes, within a single caller body.

> If that is not the case the `0` vector is checked instead if the method is not called in active path.

I feel I don't understand, because I don't see the issue. If a call is not active, then the validation result is not important anyway.

> I could do one of three things: (...)

What are the consequences of each implementation?
How is case 2 and 3 different? Again, I don't understand - if no active call, there is nothing to validate.

As I said before, the least WTF-inducing option seems to be option 2 (Or 3? Don't understand the difference).

qbojj:
> I feel I don't understand, because I don't see the issue. If a call is not active, then the validation result is not important anyway.

But if `0` vector is not a valid argument, than the transaction will not run:
```py
m = TModule()

method = Method(i=[("value", 1)])
@def_method(m, method, validate_arguments=lambda value: value == 1)
def _(value):
    pass

with Transaction().body(m):
    with m.If(self.sel):
        self.method_inner(m, value=1)
```
Here the transaction will not run if `sel==0`. This is probably a bug

tilk:
Yeah, it's a bug.

qbojj:
>  What are the consequences of each implementation?

Currently in coreblocks, all three options are equivalent

From now on:
- [1] - all-paths valid
- [2] - running paths valid
- [3] - running paths valid or 0 valid if all paths calling a method do not run

I deem [3] as a bug, so I will not add it to the considerations below.

Some things about the choice:
- expressiveness:
  - if [1] is chosen [2] can be emulated using `condition`
  - if [2] is chosen [1] cannot be easily emulated
- critical path - probably the same
- size:
  - [1] - need to have a argument checker for all paths
  - [2] - possible single argument checker per transaction

tilk:
What about implementation complexity, or the risk of combinational feedback?

I don't buy the expressiveness argument - do you have any sensible use case of [1]? I don't see it. I see the behavior of [1] as a weird corner case, not something useful.

Again, I believe [2] is more natural semantics, and we should only implement [1] if there are important reasons to do otherwise.

qbojj:
I think [2] is most reasonable.

The implementation complexity is about the same, and I don't see any scenario that would cause combinatorial feedback in either mode that could not be already done

lekcyjna123:
I agree that running path valid is the most useful semantic. If I think correctly in simple case it can be implemented by collecting all "If" conditions and-ing them and or-ing its negation with the validation result. In diamond call it also should scale well, but with more complex expressions. The only issue which I see is, what happen if an `If` depends on Method output?

For example the following scenario:
```py
m = TModule()

method = Method(i=[("value", 1)], o=[("value",1)])
@def_method(m, method)
def _(value):
    return {"value": 1}

method2 = Method(i=[("value", 1)])
@def_method(m, method2, validate_arguments=lambda value: value == 1)
def _(value):
    pass

method3 = Method(i=[("value", 1)], o=[("value",1)])
@def_method(m, method3)
def _(value):
    return {"value": ~value}

out=Signal()
out3 = Signal()
with Transaction().body(m):
    m.d.comb += out.eq(self.method(m, value=1).value)
    m.d.comb += out3.eq(self.method(m, value=out).value)
    with m.If(out)
       self.method2(m, value=out3)
```

If the transaction has been called, then `self.method` is called, then `out=1`, so `self.method3` is called, which get `out3=0`, and that blocks the `method2` so the transaction couldn't be called (is not ready). So it is not called, but this cause that `out=0` so the input to the `method2` is not validated, so it is ready, so all methods are `ready`, so transaction can be called.
Isn't this a combinational cycle? Or I am missing something?

qbojj:
The same problem can be done without any `If` statements. As such all proposed semantics have this problem. Even using a simple `m.d.comb` can cause this problem:
```py
m = TModule()

method = Method(i=[("value", 1)], validate_arguments=lambda value: value == 1)
@def_method(m, method)
def _(value):
    pass

out = Signal()
with Transaction().body(m):
    m.d.comb += out.eq(1)
    method(m, value=~out)
```

Similar problem is when `validate_arguments` argument (or `If` guarding a call to `validate_arguments` method) is defined using method arguments.
```py
m = TModule()

method_v = Method(i=[("value", 1)], validate_arguments=lambda value: value == 1)
@def_method(m, method)
def _(value):
    pass

method = Method(i=[("value", 1)])
@def_method(m, method)
def _(value):
    method_v(m, value=~value)

with Transaction().body(m):
    method(m, 1)

# make second call to prevent direct attachment of method call
with Transaction().body(m, ready=0):
    method(m, 1)
```
nothing is run -> method_v must be valid, but method doesn't have the argument (value=0) -> valid
transaction is run -> method_v has valid=1 -> invalid argument

transactron/core/manager.py
```diff
@@ -29,19 +29,32 @@
...
+    return path1[common_prefix_len].exclusive_with(path2[common_prefix_len])
+
+
 class MethodMap:
     def __init__(self, transactions: Iterable[Transaction]):
         self.methods_by_transaction = dict[TBody, list[MBody]]()
         self.transactions_by_method = defaultdict[MBody, list[TBody]](list)
-        self.argument_by_call = dict[tuple[TBody, MBody], MethodStruct]()
-        self.ancestors_by_call = dict[tuple[TBody, MBody], tuple[MBody, ...]]()
+        self.argument_by_call = defaultdict[tuple[TBody, MBody], list[tuple[ValueLike, MethodStruct]]](list)
```
tilk: The first element of the tuples is not used anywhere.

transactron/core/manager.py
```diff
@@ -447,28 +475,40 @@ def elaborate(self, platform):
...
+
         for transaction in method_map.transactions:
+
+            def validate_args_for_method(method: MBody):
+                arg_rec = Signal.like(method.data_in)
+                en = Signal()
+
+                # Only one instance of the method can be active, enables is in one-hot encoding
+                for i in OneHotSwitchDynamic(m, Cat(*method_enables[transaction][method])):
+                    m.d.comb += arg_rec.eq(method_map.argument_by_call[(transaction, method)][i][1])
```
tilk: Here you are basically zipping the arguments from `argument_by_call` with the enables from `method_enables`. It is valid because the recursion pattern in `method_enables` is the same as in `MethodMap`? If so, this is absolutely not clear and prone to breaking. Please redesign this.

transactron/core/manager.py
```diff
@@ -60,29 +73,46 @@ def report_bad_case(transaction: TBody, method: MBody, second_ancestors: tuple[M
...
-                self.ancestors_by_call[(transaction, method)] = new_ancestors
-                rec(transaction, method, new_ancestors)
+                for call_ctrl_path, arg_rec, enable_sig in calls:
+                    new_ancestors = (method, *ancestors)
+                    new_call_path = (*call_path, call_ctrl_path)
+                    if method in ancestors:
+                        report_bad_case(transaction, method, new_ancestors, new_ancestors)
+
+                    old_paths = self.call_paths_by_call[(transaction, method)]
+                    old_ancestors = self.ancestors_by_call[(transaction, method)]
```
tilk: The naming suggests that `old_ancestors` is of the same type as `new_ancestors`, but it certainly is not.
By the way, `old_paths` and `old_ancestors` are used only in one place, why not inline? Assigning them to variables activates my mutation fear - these are lists we append to!

transactron/core/manager.py
```diff
@@ -60,29 +73,46 @@ def report_bad_case(transaction: TBody, method: MBody, second_ancestors: tuple[M
...
+                    for i, old_path in enumerate(old_paths):
+                        if not call_paths_exclusive(old_path, new_call_path):
+                            report_bad_case(transaction, method, old_ancestors[i], new_ancestors)
+
+                    self.ancestors_by_call[(transaction, method)].append(new_ancestors)
+                    self.call_paths_by_call[(transaction, method)].append(new_call_path)
+                    self.argument_by_call[(transaction, method)].append((enable_sig, arg_rec))
```
tilk: We now have three things related to calls, grouped in three different lists, which share indexes by construction. And there is seemingly a fourth thing, calculated in `_method_enables`. Maybe it's time to group them together in a frozen dataclass, `CallInfo` or something?

tilk (commented):
This is rapidly changing into a significant core refactor, but it has to be done. The logic of the core needs to be as clear as possible, or we risk more weird bugs later.

transactron/core/manager.py
```diff
@@ -451,29 +473,40 @@ def elaborate(self, platform):
...
-            granted = Cat(transaction.run & method_enables[transaction][method] for transaction in transactions)
-            m.d.comb += method.run.eq(granted.any())
-
         (method_args, method_runs) = self._method_calls(m, method_map)
 
         for method in method_map.methods:
             if method.single_caller and len(method_args[method]) > 1:
                 raise RuntimeError(f"Single-caller method '{method.name}' {method.src_loc} called more than once")
             runs = Cat(method_runs[method])
+            m.d.comb += method.run.eq(runs.any())
```
tilk: This changes a lot how the method run signals are calculated. Arguably it's simpler, but this can impact performance. Is this change really needed for this feature? Did you run synthesis benchmarks on Coreblocks after this change?

qbojj: > Is this change really needed for this feature?

no, I will leave this to a different PR

transactron/core/manager.py
```diff
@@ -451,23 +491,39 @@ def elaborate(self, platform):
...
+                validate_args_for_method(method) for method in method_map.methods_by_transaction[transaction]
             ]
             runnable_terms.extend(
                 dep.run for body in method_map.ready_for_transaction(transaction) for dep in ready_dependencies[body]
             )
             m.d.comb += transaction.runnable.eq(Cat(runnable_terms).all())
 
-        ccs = _graph_ccs(cgr)
-
         method_enables = self._method_enables(method_map)
```
tilk: Huh, wasn't `_method_enables` now redundant because of `call_info`?

tilk:
Coreblocks does not synthesize, combinational cycle is created. I think that the culprit may be the removed special case for only 1 call, which was in `_set_method_uses`.

tilk:
By the way, it would be nice to have a Verilog export test for corner cases like this.

qbojj:
The problem path passes through the 'en' signal of the argument checker.

qbojj:
I thought amaranth was doing constant folding for operations before generating the code - it does not

tilk (approved):
Looks like Coreblocks synthesizes now and the performance is not impacted.

PR #146, opened 2026-04-20
Probably actually fix method readiness (maybe)

### LINKED ISSUE #145: Nested transaction in method with more than one caller not being called when only one caller is running
Sorry I think I spoke too soon for the fix in https://github.com/kuznia-rdzeni/transactron/pull/133. I swear my tests used to pass with the patch but it started failing as soon as I upgrated to 0.6.0 and it still didn't work when I switched back to my previous patched version on top of 0.5.1.

The code reproducing the issue can be seen at the bottom.

Both top-level transactions are calling the same `method` which contains a nested transaction that's always ready and does nothing more than printing a message. As long as one of the outer transactions are running the method should be called and the inner transaction should be running.

However, the actual printing from the code is,

```
Method runing, trans.run: 0
Outer transaction 1
Method runing, trans.run: 0
Outer transaction 1
Method runing, trans.run: 0
Outer transaction 1
Inner transaction running
Method runing, trans.run: 1
Outer transaction 1
Inner transaction running
Method runing, trans.run: 1
Outer transaction 1
Inner transaction running
Method runing, trans.run: 1
Outer transaction 1
```

The method is always being called no matter whether one or both of the outer transactions are ready, however, the inner transaction (within the method) is only running when both outer transactions are ready (even though only one of them will be running).

Removing one of the caller fixes the issue (which probably corresponds to the test I originally did for #133 even though I thought I did more...). Adding other ready flags, making the method nonexclusive does not make a difference. Based on output from custom combiner, this seems to only happen when there's more than one possible caller. Maybe some signal got cross-wired?

```python
from amaranth import *
from amaranth.lib.wiring import Component, In, Out

from transactron import TModule, Transaction, Method, def_method

class Outer(Component):
    I: In(1)

    def elaborate(self, _):
        m = TModule()

        method = Method()
        @def_method(m, method)
        def _():
            trans = Transaction()
            with trans.body(m):
                m.d.sync += Print("Inner transaction running")
            m.d.sync += Print("Method runing, trans.run:", trans.run)

        with Transaction().body(m):
            m.d.sync += Print("Outer transaction 1")
            method(m)

        with Transaction().body(m, ready=self.I):
            m.d.sync += Print("Outer transaction 2")
            method(m)

        return m

if __name__ == '__main__':
    from amaranth.sim import Simulator
    from transactron import TransactronContextElaboratable
    p = Outer()
    m = TransactronContextElaboratable(p)
    sim = Simulator(m)
    sim.add_clock(1e-6)
    async def test(ctx):
        for _ in range(3):
            await ctx.tick()
        ctx.set(p.I, 1)
        for _ in range(3):
            await ctx.tick()
    sim.add_testbench(test)
    sim.run()
```

qbojj:
Close #145 

Fixes the ready_dependent logic

### DIFF
--- a/transactron/core/manager.py
+++ b/transactron/core/manager.py
@@ -245,8 +245,8 @@ class TransactionManager(Elaboratable):
         return method_enables
 
     @staticmethod
-    def _ready_dependencies(method_map: MethodMap) -> Graph[Body]:
-        ready_dependencies = defaultdict[Body, set[Body]](set)
+    def _ready_dependencies(method_map: MethodMap) -> Mapping[TBody, set[Body]]:
+        ready_dependencies = defaultdict[TBody, set[Body]](set)
 
         relations = TransactionManager._relations(method_map)
 
@@ -254,14 +254,8 @@ class TransactionManager(Elaboratable):
             if not relation.ready_dependent:
                 continue
 
-            # Transitively trickle down ready dependencies
-            todo = {relation.end}
-            while todo:
-                current = todo.pop()
-                if relation.start not in ready_dependencies[current]:
-                    ready_dependencies[current].add(relation.start)
-                    for parent in method_map.method_parents[current]:
-                        todo.add(parent)
+            for trans in method_map.transactions_for(relation.end):
+                ready_dependencies[trans].add(relation.start)
 
         return ready_dependencies
 
--- a/transactron/core/manager.py
+++ b/transactron/core/manager.py
@@ -112,7 +112,7 @@ class TransactionManager(Elaboratable):
         self.cc_scheduler = cc_scheduler
 
     @staticmethod
-    def _relations(method_map: MethodMap) -> Sequence[Relation[Body]]:
+    def _relations(method_map: MethodMap) -> Sequence[Relation]:
         return [
             Relation(start=elem, **dataclass_asdict(relation))
             for elem in method_map.methods_and_transactions
--- a/transactron/core/manager.py
+++ b/transactron/core/manager.py
@@ -245,8 +245,8 @@ class TransactionManager(Elaboratable):
         return method_enables
 
     @staticmethod
-    def _ready_dependencies(method_map: MethodMap) -> TransactionGraph:
-        ready_dependencies = defaultdict[TBody, set[TBody]](set)
+    def _ready_dependencies(method_map: MethodMap) -> Graph[Body]:
+        ready_dependencies = defaultdict[Body, set[Body]](set)
 
         relations = TransactionManager._relations(method_map)
 
@@ -254,12 +254,14 @@ class TransactionManager(Elaboratable):
             if not relation.ready_dependent:
                 continue
 
-            for trans_start in method_map.transactions_for(relation.start):
-                for trans_end in method_map.transactions_for(relation.end):
-                    if trans_start is not trans_end and not TransactionManager._transactions_exclusive(
-                        method_map, trans_start, trans_end
-                    ):
-                        ready_dependencies[trans_end].add(trans_start)
+            # Transitively trickle down ready dependencies
+            todo = {relation.end}
+            while todo:
+                current = todo.pop()
+                if relation.start not in ready_dependencies[current]:
+                    ready_dependencies[current].add(relation.start)
+                    for parent in method_map.method_parents[current]:
+                        todo.add(parent)
 
         return ready_dependencies
 
@@ -452,7 +454,7 @@ class TransactionManager(Elaboratable):
                 method._validate_arguments(method_map.argument_by_call[transaction, method])
                 for method in method_map.methods_by_transaction[transaction]
             ]
-            runnable_terms.extend(dep.ready for dep in ready_dependencies[transaction])
+            runnable_terms.extend(dep.run for dep in ready_dependencies[transaction])
             m.d.comb += transaction.runnable.eq(Cat(runnable_terms).all())
 
         ccs = _graph_ccs(cgr)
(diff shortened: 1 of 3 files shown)


transactron/core/manager.py
```diff
@@ -452,7 +454,7 @@ def elaborate(self, platform):
                 method._validate_arguments(method_map.argument_by_call[transaction, method])
                 for method in method_map.methods_by_transaction[transaction]
             ]
-            runnable_terms.extend(dep.ready for dep in ready_dependencies[transaction])
+            runnable_terms.extend(dep.run for dep in ready_dependencies[transaction])
```
tilk: Huh, did I really miss that? 👀

transactron/core/manager.py
```diff
@@ -245,21 +245,23 @@ def rec(transaction: TBody, source: Body):
...
             if not relation.ready_dependent:
                 continue
 
-            for trans_start in method_map.transactions_for(relation.start):
-                for trans_end in method_map.transactions_for(relation.end):
-                    if trans_start is not trans_end and not TransactionManager._transactions_exclusive(
-                        method_map, trans_start, trans_end
-                    ):
-                        ready_dependencies[trans_end].add(trans_start)
+            # Transitively trickle down ready dependencies
```
tilk: Isn't the transitivity implicit?

tilk (commented):
Tests missing.

transactron/core/manager.py
qbojj: I'm not sure if I understand

yuyichao:
I did some fuzz tests with random nested Method/Transaction and call graphs and for all the non-cyclic ones I've checked the behavior is the same as my expectation.

There are some cases where two methods effectively call each other that I've seen the `run` signal doesn't match whether the body actually get run or not but I assume it doesn't matter much.

qbojj:
@yuyichao Could you send the tests that have the `run` signal not matching.

yuyichao:
Sure,

```python
from amaranth import *
from amaranth.lib.wiring import Component, In, Out

from transactron import TModule, Transaction, Method, def_method

class Outer(Component):
    ready: In(3)
    run1: Out(3)
    run2: Out(3)

    def elaborate(self, _):
        m = TModule()

        method1 = Method()
        method2 = Method()
        trans3 = Transaction()

        m.d.top_comb += self.run1[0].eq(method1.run)
        m.d.top_comb += self.run1[1].eq(method2.run)
        m.d.top_comb += self.run1[2].eq(trans3.run)

        @def_method(m, method1, nonexclusive=True)
        def _():
            m.d.comb += self.run2[0].eq(1)
            method2(m)
            @def_method(m, method2, ready=self.ready[1])
            def _():
                m.d.comb += self.run2[1].eq(1)
                method1(m)
                with trans3.body(m, ready=self.ready[2]):
                    m.d.comb += self.run2[2].eq(1)

        return m

if __name__ == '__main__':
    from amaranth.sim import Simulator
    from transactron import TransactronContextElaboratable
    p = Outer()
    m = TransactronContextElaboratable(p)

    sim = Simulator(m)
    # sim.add_clock(1e-6)
    async def test(ctx):
        ctx.set(p.ready, 4)
        print(ctx.get(p.run1), ctx.get(p.run2))
    sim.add_testbench(test)
    sim.run()
```

As I said, it contains a cycle of method call/nesting so I'm not sure if this is valid code.
However, it does not error and prints `4 0` suggesting that the run signal thinks trans3 would be running even tough it's not actually running.

qbojj:
Turns out relations from uncalled methods get pruned, so the ready dependency is also removed

yuyichao:
Ah, interesting.

It turns out that the way I randomly generate the graph would actually only create uncalled methods if there's a cycle and that's why the non-cyclic graphs are all passing the test. After tweaking the generation code I'm seeing a lot of failures now. From a rough glance, it does seem that they all contain uncalled methods.

tilk:
> Turns out relations from uncalled methods get pruned, so the ready dependency is also removed

Damn it, such a weird corner case. I think that pruning related to uncalled methods was probably a mistake. Defined methods make much more sense, but they were introduced much later; originally methods were gathered by following call chains from transactions, which were registered directly into the transaction manager, and the transaction and method keys did not exist yet. Feel free to clean up historical cruft if needed.

Edit: This PR seems pretty clean other than the comment below, I would leave cleaning up to future PRs and merge this soon.

transactron/core/manager.py
```diff
@@ -449,10 +444,13 @@ def elaborate(self, platform):
...
             runnable_terms = [
-                method._validate_arguments(method_map.argument_by_call[transaction, method])
+                (
+                    method._validate_arguments(method_map.argument_by_call[transaction, method])
+                    & Cat(dep.run for dep in ready_dependencies[method]).all()
+                )
                 for method in method_map.methods_by_transaction[transaction]
             ]
-            runnable_terms.extend(dep.ready for dep in ready_dependencies[transaction])
+            runnable_terms.extend(dep.run for dep in ready_dependencies[transaction])
```
tilk: I don't think I like that the `ready_dependencies` logic is in two places now, and one of the places is mixed up with `validate_arguments` logic. How about something like:

```suggestion
            runnable_terms.extend(dep.run for body in [transaction] + method_map.methods_by_transaction[transaction] for dep in ready_dependencies[body])
```

But aaaactually, I see that "transaction and its called methods" thingy is already used in `_transactions_exclusive`, two times. This looks like a new function on `method_map`.

transactron/core/manager.py
tilk: Brain fart, please ignore.

yuyichao:
Local tests passed for me for the current version.

tilk:
@yuyichao Thanks for the testing!

tilk (approved):

PR #147, opened 2026-04-20
fix nonexclusive check

qbojj:
fixes the check for presence of combiner in `nonexclusive` methods

### DIFF
--- a/transactron/lib/transformers.py
+++ b/transactron/lib/transformers.py
@@ -1,7 +1,6 @@
 from amaranth import *
 from amaranth_types import ValueLike, ModuleLike, HasElaborate
 
-from transactron.utils.amaranth_ext.elaboratables import OneHotSwitchDynamic
 from transactron.utils.transactron_helpers import get_src_loc
 from ..core import *
 from ..utils import SrcLoc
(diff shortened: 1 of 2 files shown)


transactron/lib/transformers.py
```diff
@@ -530,7 +531,16 @@ def create(target: Method, *, src_loc: int | SrcLoc = 0):
...
 
-        @def_method(m, self.method, nonexclusive=True)
+        def combiner(m, args, runs):
+            if len(args) == 1:
+                return args[0]
+            else:
+                arg = Signal.like(self.method.data_in)
+                for i in OneHotSwitchDynamic(m, runs):
+                    m.d.comb += arg.eq(args[i])
+                return arg
```
tilk: Is this a copy of `default_combiner`?

tilk (approved):

yuyichao:
Hmm, so is the intention that basically all nonexclusive method should supply a custom combiner?

yuyichao:
It'll be nice if the default combiner can be exposed so that it can be used for cases where I know there'll be only one caller (basically the case `NonexclusiveWrapper` is for but I already know that for internal modules where I know how it'll be used and don't want to create unnecessary wrapper).

PR #148, opened 2026-04-27
allow for multiple calls to nonexclusive methods

### LINKED ISSUE #129: Only one call to a particular nonexclusive method is allowed
```py
m = TModule()
method1 = Method()

@def_method(m, method1, nonexclusive=True)
def _():
    pass

with Transaction():
    method1(m)
    method1(m)
```

is not allowed, even though it would be well defined. The only argument that I could find against this is that it would be a bit easier to create unstable circuits using transactron (while using combiners - should not happen for output-only nonexclusive functions).

qbojj:
Closes #129

It also adds checking on uncalled methods, while also unifying exclusive path handling


transactron/core/manager.py
```diff
@@ -80,6 +77,28 @@ def report_bad_case(
 
             raise RuntimeError(msg)
 
+        def validate_root_call_tree(root: Body):
+            info_by_call = defaultdict[MBody, list[tuple[tuple[MBody, ...], tuple[CtrlPath, ...]]]](list)
```
tilk: Use a different name, this is confusing.

transactron/core/manager.py
```diff
@@ -47,7 +47,7 @@ class CallInfo:
 
 
 class MethodMap:
-    def __init__(self, transactions: Iterable[Transaction]):
+    def __init__(self, transactions: Iterable[Transaction], methods: Iterable[Method]):
         self.methods_by_transaction = dict[TBody, list[MBody]]()
         self.transactions_by_method = defaultdict[MBody, list[TBody]](list)
```
tilk: You pass the set of all defined methods. While you do this refactor, does it make sense to change this to a normal `dict` and initialize with the set of methods? This way `MethodMap.methods()` will start to report the full set of defined methods, which could make the core logic healthier. Hopefully nothing breaks after this change :)

test/core/test_methods.py
```diff
@@ -203,6 +203,56 @@ def elaborate(self, platform):
 
         self.assert_re("called twice", Twice())
 
+    def test_twice_nonexclusive(self):
```
tilk, line 207: A test with a unifier which verifies that unifiers work in this case would be nice, too.

transactron/core/manager.py
```diff
@@ -80,6 +77,28 @@ def report_bad_case(
...
+
+            def rec_root(source: Body, ancestors: tuple[MBody, ...], call_path: tuple[CtrlPath, ...]):
+                for method_obj, calls in source.method_calls.items():
+                    method = MBody(method_obj._body)
+                    for call_ctrl_path, _, _ in calls:
+                        new_ancestors = (method, *ancestors)
+                        new_call_path = (*call_path, call_ctrl_path)
+
+                        if method in ancestors:
+                            report_bad_case(root, method, new_ancestors, new_ancestors)
```
tilk: Maybe separate reporting for double calls and cycles? There is a weird `if` in `report_bad_case` that would go away.

tilk: Maybe use a last common ancestor instead of `root` for error reporting, the message will then be cleaner.

tilk (changes requested):

transactron/core/manager.py
qbojj: sounds good

transactron/core/manager.py
```diff
@@ -507,6 +528,8 @@ def validate_args_for_method(method: MBody):
         for method in method_map.methods:
```
tilk: How about another attribute/property on `MethodMap` for called methods (the previous behavior)?

qbojj:
Now that I think about it, we could also allow for nonexclusive methods to call themselves

tilk:
There is still the failing test.

qbojj:
:3

tilk:
Is this useful for anything other than creating combinational loops?

qbojj:
self-calling may be useful for example:
- we have global state `S` with nonexclusive `read`
- the `read` calls `read2` to get `S2` - a part of the state
- `S2` has an invariant against the value of `S`
- `read2` calls `read` just to do assertion checks

tilk:
> self-calling may be useful for example:

Well, I agree that there is no sense in introducing restrictions which don't serve any purpose. Still, this PR already has a lot of changes.

I have verified that the current state of this PR works with Coreblocks. Time to merge this.

tilk (approved):

PR #149, opened 2026-04-27
Remove duplicated state from Body

qbojj:
changes `conditional_calls` from member variable to property, as `method_calls` already has (almost) all of the required information.

It also changes enable_call to be a sugar around `with m.If(enable_call): ...`


transactron/core/body.py
```diff
@@ -70,11 +69,18 @@ def __init__(
             kwargs["validate_arguments"] if "validate_arguments" in kwargs else None
         )
         self.method_calls = defaultdict(list)
-        self.conditional_calls = set()
 
         if self.nonexclusive:
             assert len(self.data_in.as_value()) == 0 or "combiner" in kwargs
 
+    @property
```
tilk:
```suggestion
    @cached_property
```

tilk: Or actually, I don't have a good idea right now. Transactron elaboration works in stages:

- First the source is elaborated,
- Then simultaneous transactions are elaborated,
- Then the set of transactions, methods etc. is considered as final and should not change.

Making this a cached property has the risk that this will be touched before the final stage, causing bugs. But not making this a cached property causes a possible performance regression (this is called under a triple for loop).

qbojj: I think the best way would be to just change it to a function (to show that this is not O(1) operation) and just build a cache in `_conditionally_called_methods`

tilk: This is used in one place, so maybe it's fine, on the other fine it would add a needless complication to a function which now looks simple.

The core of Transactron should be as declarative as possible. We should not fear duplicating things for simplicity or performance as long as there is no risk of things getting out of sync - and after simultaneous transactions, there is no risk. Your change in principle is a good one - it reduces the mutable part of the system - but I think the memoization should remain.

tilk (approved):

PR #151, opened 2026-04-27
Circular allocator

tilk:
Sometimes a FIFO-like allocation system is needed without an actual FIFO. This PR factors out the allocator of `BasicFifo` and extends it with multi-allocation.

TODO:
- [x] Return all allocated/deallocated IDs for convenience
- [x] Docstrings
- [x] Tests

### DIFF
--- a/test/lib/test_allocators.py
+++ b/test/lib/test_allocators.py
@@ -98,3 +98,68 @@ class TestPreservedOrderAllocator(TestCaseWithSimulator):
             sim.add_testbench(order_verifier, background=True)
             sim.add_testbench(allocator)
             sim.add_testbench(deallocator)
+
+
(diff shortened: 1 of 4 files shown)


awariac:
> Return all allocated/deallocated IDs for convenience

Would be helpful to add extra option to enable logging of each allocation/deallocation, as similar mechanism is used e.g. in Coreblocks CRAT.

tilk:
I think that it's probably better not to include logging here. Logging messages are nicer when more domain-specific information is available, and it always can be added at the use site.

I'm experimenting with the docstring format for methods. I think that it's better to use shapes rather than types after the colons, as this is the information you actually need when interfacing with a method.

qbojj (approved):

PR #156, opened 2026-04-30
HDL logging wrappers

awariac:
Module wrappers (similar to Transacti?Component) to enable HDL logging backend, e.g. synthesized to Verilog code (very useful!).

Conversion to Framgent is performed first, because it needs to force elaboration and add all log records to LogKey first.


tilk:
This is for use with other toolchains when integrating Coreblocks into larger systems?

tilk:
Are you aware of any way to test this? I'm reluctant to include code not covered by any tests.

awariac:
ok, found a way

awariac:
> This is for use with other toolchains when integrating Coreblocks into larger systems?

yes, it was useful when debugging Linux boot on Litex, using external Verilator simulation run

awariac:
requires https://github.com/kuznia-rdzeni/amaranth-stubs/pull/14

transactron/testing/logging.py
```diff
@@ -107,3 +115,83 @@ async def log_process(sim: ProcessContext) -> None:
...
+    def elaborate(self, platform):
+        m = Module()
+
+        elaboratable = Fragment.get(self.elaboratable, platform)
+        m.submodules.elaboratable = elaboratable
+
+        cycle = Signal(64)
+        m.d.sync += cycle.eq(cycle + 1)
+        if self.print_cycle_separator:
+            m.d.sync += Print(Format("--- CYCLE {} ---", cycle))
```
qbojj: It may be better if the CYCLE header would be outputted only if some trigger is fired

qbojj (approved):

tilk:
> requires [kuznia-rdzeni/amaranth-stubs#14](https://github.com/kuznia-rdzeni/amaranth-stubs/pull/14)

Merged and made a new release.

awariac:
@tilk should stubs be updated to >= requirement instead of ~=?
Repository won't type check otherwise 0.1.3, but it will work. Maybe we should have a strong requirement only in [dev]?

tilk:
Why? `~=` is the compatible release operator, which is exactly what is needed for semantic versioning to work.

Actually the problem is your changes in `amaranth-stubs` were not backwards-compatible and I used a wrong version number. Thought you checked that the changes don't break Transactron and Coreblocks, but looks like I was wrong. Please do the checks next time.

tilk:
Re-published the new stubs as `0.2.0` and yanked `0.1.3`.

Also, I'm thinking of changing the versioning to follow `amaranth` itself, and add another two numbers for the stubs (so that incompatible stub releases can still be done for a given Amaranth version). I.e. `A.A.A.S.S`, where `A` is from Amaranth, and `S` is from stubs.

awariac:
would `A.A.S.S` be enough? Amaranth doesn't do removals on 0.x.{1+} versions.

tilk:
> would `A.A.S.S` be enough? Amaranth doesn't do removals on 0.x.{1+} versions.

Right, `A.A.S.S` should be enough.

tilk:
@awariac: I have made two other changes to `amaranth-stubs`, which make both this code and Coreblocks typecheck correctly again. If you confirm, I will release `amaranth-stubs` version 0.5.0.0, to match the agreed versioning scheme.

awariac:
looks like it works

pyproject.toml
```diff
@@ -9,7 +9,7 @@ readme = "README.md"
 description = "Hardware transactions library for Amaranth"
 dependencies = [
     "amaranth ~= 0.5.8",
-    "amaranth-stubs ~= 0.1.2",
+    "amaranth-stubs ~= 0.5.0.0"
```
tilk:
```suggestion
    "amaranth-stubs ~= 0.5.0.0",
```

tilk (approved):

PR #157, opened 2026-05-01
Remove all Record references

awariac:
`Record`s were deprecated long time ago and are removed in Amaranth `0.6.0`.

idk if the new internal type names look fine

This can be merged in when Amarath 0.6.0 gets released or now


tilk:
The renamings will need to be done in Coreblocks, too, but I agree - records are deprecated, we don't use them anymore, it's time to let them go.

tilk (approved):

PR #159, opened 2026-05-02
Logging system refactor

tilk:
The logging system was designed before some large changes in Amaranth (including new testing system, and possibly ValueCastables etc.). Given recent developments in Coreblocks (kuznia-rdzeni/coreblocks#934), it would be nice if the logging system was a little more generic.

The main goal is to allow `Value`s and `ValueCastable`s in logging. This would allow module-less usage of logging functions, and possibly nicer output for `ValueCastable`s. Pysim works immediately, but use with Verilator etc. will probably require some wrapping.

TODO:
- [x] Allow `Value` and `ValueCastable` in `LogRecord`
- [x] Rename `lib.logging` to `utils.logging` because that's where it conceptually should be in the first place. The intended order of dependencies is `utils` -> `core` -> `lib`.
- [x] Make cocotb/verilator interoperability work again
- [x] Add module-less logging API
- [x] Tests


transactron/utils/logging.py
```diff
@@ -191,7 +251,27 @@ def assertion(self, m: ModuleLike, value: Value, format: str = "", *args, src_lo
 
         See `HardwareLogger.log` function for more details.
         """
-        self.error(m, ~value, format, *args, **kwargs, src_loc_at=src_loc_at + 1)
+        self.error(m, ~Value.cast(value), format, *args, src_loc=get_src_loc(src_loc))
```
qbojj: As the assertion value can have more than one bit

```suggestion
        self.error(m, ~Value.cast(value).any(), format, *args, src_loc=get_src_loc(src_loc))
```

transactron/utils/logging.py
```diff
@@ -109,7 +109,85 @@ def __init__(self, name: str):
...
+
+    def top_assertion(self, value: ValueLike, format: str, *args: ValueLike, src_loc: int | SrcLoc = 0):
+        """Define an assertion.
+
+        Unlike `HardwareLogger.assertion`, this function can be used in
+        contexts where a module is not available.
+
+        See `HardwareLogger.assertion` function for more details.
+        """
+        self.top_error(~Value.cast(value), format, *args, src_loc=get_src_loc(src_loc))
```
qbojj: same as `assertion`: `~Value.cast(value).any()`

transactron/utils/logging.py
```diff
@@ -120,67 +198,49 @@ def log(self, m: ModuleLike, level: LogLevel, trigger: ValueLike, format: str, *
...
             How many stack frames below to look for the source location, used to
             identify the failing assertion.
         """
-
-        def local_src_loc(src_loc: SrcLoc):
-            return (os.path.relpath(src_loc[0]), src_loc[1])
-
-        src_loc = local_src_loc(get_src_loc(src_loc_at + 1))
-
         trigger_signal = Signal()
```
qbojj: trigger could have more than 1 bit

```suggestion
        trigger_signal = Signal.like(trigger)
```

qbojj (approved):
Other than those three I see no issues

transactron/utils/logging.py
tilk: Good catch.

PR #162, opened 2026-05-04
Bugfix for validate_arguments

tilk:
Huh, looks like I introduced a small bug.


qbojj (approved):
Just noticed it when looking at lately merged PRs too

PR #164, opened 2026-05-04
Add def_method_mocks

### LINKED ISSUE #154: Add def_method_mocks
As superscalarity support improves, it would be useful to have a simple way to mock `Methods` in simulation.

tilk:
Work-in-progress PR to add `def_method_mocks`.

Fixes #154

### DIFF
--- a/transactron/testing/method_mock.py
+++ b/transactron/testing/method_mock.py
@@ -1,6 +1,6 @@
 from contextlib import contextmanager
 import functools
-from typing import Callable, Any, Optional, Unpack
+from typing import Callable, Any, Optional, Unpack, NotRequired
 
 from amaranth.sim._async import SimulatorContext
 from transactron.core.body import AdapterBodyParams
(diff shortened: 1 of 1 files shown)


transactron/testing/method_mock.py
```diff
@@ -13,27 +13,30 @@
...
+            assert "validate_arguments" not in kwargs
+            assert abp == {}
         self.adapter = adapter
         self.function = function
-        self.validate_arguments = validate_arguments
-        self.enable = enable
-        self.delay = delay
+        self.validate_arguments = kwargs["validate_arguments"] if "validate_arguments" in kwargs else None
+        self.enable = kwargs["enable"] if "enable" in kwargs else lambda: True
+        self.delay = kwargs["delay"] if "delay" in kwargs else 0
```
Arusekk, line 39:
```suggestion
        self.validate_arguments = kwargs.get("validate_arguments")
        self.enable = kwargs.get("enable", lambda: True)
        self.delay = kwargs.get("delay", 0)
```

I also hope that we can keep the argument syntax instead of parsing dicts, so that 'dumber' tooling can still give valuable hints. Or is it a typing limitation we are working around with code here?

PR #168, opened 2026-05-11
Testing dependencies

tilk:
This PR moves `pytest` and `hypothesis` to dependencies. This is the simplest solution of the optional dependency problem for the `testing` submodule.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,9 @@ dependencies = [
     "amaranth-stubs ~= 0.5.0.0",
     "dataclasses-json ~= 0.6.3",
     "tabulate ~= 0.9.0",
-    "networkx ~= 3.4.2"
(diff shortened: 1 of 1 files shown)


awariac (approved):
Ok for now, I see why it was not included in the first place -

Some other idea:
Those dependencies could be in a separate [testing-extras] dependency group, could that uses them moved to `transactron.testing.extras` sub-package.
We can then raise an error in `testing/extras/__init__.py` based on `importlib.metadata.reaquirements` if dependencies are not installed and someone tries to use `transactron.testing.extras`.
I'm also not sure if `pytest` infrastructure utilities are core-enough or not to be included there. 


\+ I'm worried about commonly used `pytest` requirement that may be too strict.
Current release is `9.x`. We may test if our helpers work with it and include it in `>=, <` requirements

PR #170, opened 2026-05-11
Fix UniferKey, add DependencyKey tests

awariac:
Fixes #169


transactron/lib/dependencies.py
```diff
@@ -1,4 +1,5 @@
 from collections.abc import Callable
+from typing import Iterable
```
tilk:
```suggestion
from collections.abc import Callable, Iterable
```

test/lib/test_dependencies.py
```diff
@@ -0,0 +1,85 @@
...
+
+        with self.run_simulation(m) as sim:
+            sim.add_testbench(test_process)
+
+    @pytest.mark.parametrize("key", [ListKeyInstance(), ListKeyInstanceNoLockEmptyValid()])
+    def test_list_key(self, key):
+        dm = DependencyContext.get()
+
+        if isinstance(key, ListKeyInstanceNoLockEmptyValid):
+            assert dm.get_dependency(key) == []
```
tilk: `empty_valid` is the default on list keys.

test/lib/test_dependencies.py
```diff
@@ -0,0 +1,85 @@
...
+        def elaborate(self, platform):
+            m = TModule()
+
+            method, unifier = DependencyContext.get().get_dependency(UnifierKeyProduct())
+
+            m.submodules += unifier
+
+            self.a.provide(method)
+
+            return m
```
tilk: This is a test module for unifier tests, but isn't named correspondingly.

Also, code style remark: I'm not a fan of test circuits inside test classes, this increases indentation level for no real reason. It's not like we need to encapsulate anything or prevent namespace pollution.

test/lib/test_dependencies.py
```diff
@@ -0,0 +1,85 @@
...
+            m.c.enable(sim)
+            m.a.call_init(sim)
+            await sim.tick()
+            assert m.a.get_done(sim)
+            assert m.b.get_done(sim)
+            assert m.c.get_done(sim)
+
+        with self.run_simulation(m) as sim:
+            sim.add_testbench(test_process)
+
```
tilk: There is no `SimpleKey` test, `empty_valid` is not correctly tested, `cache` is not tested at all.

test/lib/test_dependency_key.py
```diff
@@ -0,0 +1,156 @@
...
+class KeyCache(CacheTestKey):
+    pass
+
+
+@dataclass(frozen=True)
+class UnifierKeyProduct(UnifierKey, unifier=MethodProduct.create):
+    pass
+
+
+class CircuitTestModule(Elaboratable):
```
tilk: `UnifierKeyTestModule`?

transactron/lib/dependencies.py
```diff
@@ -8,27 +9,24 @@
...
+    `Unifier` module needs to be added as submodule when calling `combine`.
     """
 
-    unifier: Callable[[list["Method"]], "Unifier"]
-
     def __init_subclass__(cls, unifier: Callable[[list["Method"]], "Unifier"], **kwargs) -> None:
-        cls.unifier = unifier
-        return super().__init_subclass__(**kwargs)
+        # drop `self` argument when assigning to cls
+        cls.unifier = lambda _, targets: unifier(targets)
```
tilk: This should be more Pythonic?

```suggestion
        cls.unifier = staticmethod(unifier)
```

test/lib/test_dependency_key.py
```diff
@@ -0,0 +1,156 @@
...
+        self.dm = DependencyContext.get()
+        self.dm.add_dependency(UnifierKeyProduct(), self.b)
+        self.dm.add_dependency(UnifierKeyProduct(), self.c)
+
+    def elaborate(self, platform):
+        m = TModule()
+
+        method, unifier = DependencyContext.get().get_dependency(UnifierKeyProduct())
+
+        m.submodules += unifier
```
qbojj, line 74: Now that I think about it, unifier keys should never be gotten more than once, as if that were the case the unifier would be added as submodule in multiple places.

This would probably need a read-locking mechanism.

tilk: They could also be non-cached, which would create a new unifier for each use site.

awariac: > They could also be non-cached, which would create a new unifier for each use site.

i was about to write that, it could be useful in some cases

will change later

tilk (approved):

PR #188, opened 2026-06-30
Allocator clear

tilk:
This PR adds `clear` methods to `PriorityEncoderAllocator` and `PreservedOrderAllocator`. Additionally, `peek` and `replace` are added to `PriorityEncoderAllocator`, in anticipation of proper checkpointing in Coreblocks.

TODO:
* [x] Tests

### DIFF
--- a/transactron/lib/allocators.py
+++ b/transactron/lib/allocators.py
@@ -16,15 +16,25 @@ class PriorityEncoderAllocator(Elaboratable):
     This module allows to allocate and deallocate identifiers from a continuous
     range. Multiple identifiers can be allocated or deallocated in a single
     clock cycle.
-
-    Attributes
-    ----------
-    alloc : Methods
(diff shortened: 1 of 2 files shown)


transactron/lib/allocators.py
```diff
@@ -63,6 +76,18 @@ def _(i):
...
 
+        @def_method(m, self.peek)
+        def _():
+            return {"mask": not_used}
+
+        @def_method(m, self.replace)
+        def _(mask):
+            m.d.sync += not_used.eq(mask)
+
+        @def_method(m, self.clear)
```
qbojj: should probably be nonexclusive

qbojj (approved):

transactron/lib/allocators.py
tilk: That was the idea but I forgot.

PR #189, opened 2026-07-02
Structured hardware event log (evlog)

xThaid:
A second take on hardware logging, but this time structured: instead of formatted text messages, the design emits typed events that are captured during simulation into an event log with a stable schema. The log can be saved to a file and processed offline by downstream tools (trace converters, statistics, whatever) without touching the HDL again.

The hardware logging module can be technically migrated to use this instead.

### DIFF
--- a/docs/evlog.md
+++ b/docs/evlog.md
@@ -1,12 +1,12 @@
 # Event log
 
-The `transactron.evlog` package instruments a hardware design with structured, typed events which are captured during simulation into an *event log*. Event logs are consumed by downstream tools - pipeline trace converters, statistics scripts, checkers - without touching the hardware description again.
+The {py:mod}`transactron.evlog` package instruments a hardware design with structured, typed events which are captured during simulation into an *event log*. Event logs are consumed by downstream tools - pipeline trace converters, statistics scripts, checkers - without touching the hardware description again.
 
 The event log carries machine-readable payloads: named, typed fields with a stable schema. The same captured log can be decoded offline, long after the simulation finished, as long as the schema (saved together with the log) and the event definitions are available.
 
 ## Defining events
 
-Events are dataclasses declared with the `event` decorator. The event name must be globally unique:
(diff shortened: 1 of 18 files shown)


docs/evlog.md
```diff
@@ -0,0 +1,107 @@
+# Event log
+
+The `transactron.evlog` package instruments a hardware design with structured, typed events which are captured during simulation into an *event log*. Event logs are consumed by downstream tools - pipeline trace converters, statistics scripts, checkers - without touching the hardware description again.
```
tilk: Could you use Sphinx references (see `getting-started.md`)? One advantage of using them (other than getting API links) is that you get warnings if code gets out-of-sync with the doc.

docs/evlog.md
```diff
@@ -0,0 +1,107 @@
...
+
+with self.run_simulation(m) as sim:
+    log, process = capture_evlog(metadata={"config": "full"})
+    sim.add_process(process)
+    sim.add_testbench(testbench)
+
+events = log.decoded()      # or log.save("events.jsonl")
+```
+
+Tests based on {py:class}`~transactron.testing.test_case.TestCaseWithSimulator` capture event logs automatically when pytest is invoked with `--transactron-evlog`: every test which registered any emission sites (i.e. tests which enabled the event log themselves) saves its captured events to `test/__evlogs__/<test name>.jsonl`. The flag only controls capturing - enabling the event log stays a separate concern.
```
qbojj: this should probably mention `__TRANSACTRON_EVLOG` rather than `--transactron-evlog`, as `--transactron-evlog` is part of conftest, while `__TRANSACTRON_EVLOG` will work everywhere

qbojj (approved):

docs/evlog.md
qbojj: Alternatively could expose reusable `transactron.testing.pytest` submodule that would expose `--transactron-*` option group with its handling.

PR #190, opened 2026-07-03
Transpose

tilk:
This PR adds the `transpose` function, which swaps two layers of `data.View`. Intended use: superscalarity in pipeline builder.

TODO:
- [x] Docs
- [x] Test


test/utils/amaranth_ext/test_data.py
qbojj, line 1: we could also test if transpose of transpose is identity

qbojj (approved):

PR #197, opened 2026-07-17
Extended assign

tilk:
This PR extends the `assign` function (which is used e.g. in method calling) with support for `data.Const` and Amaranth enums. This should allow to improve enum strictness in Coreblocks.


qbojj (approved):

transactron/utils/assign.py
```diff
@@ -136,8 +139,8 @@ def rec_call(name: str | int):
             lhs[name],  # type: ignore
             rhs[name],  # type: ignore
             fields=subfields,
-            lhs_strict=isinstance(lhs, ValueLike),
-            rhs_strict=isinstance(rhs, ValueLike),
+            lhs_strict=isinstance(lhs, ValueLike) and not isinstance(lhs[name], int),  # type: ignore
```
qbojj, line 142: Is this even possible? LHS shouldn't have any consts or ints (or at least all assigned costs agree on value assigned to them - like wiring.connect)

tilk: Not really, added mostly for better error messages.

qbojj: Than fine by me. We can merge

PR #201, opened 2026-07-19
add always-call sugar

qbojj:
Closes #198


transactron/core/method.py
```diff
@@ -332,6 +330,46 @@ def __repr__(self) -> str:
...
+                ret = method(m, *args, **kwargs)
+            assertion(m, t.run, ...)
+
+        if not in a transaction body, and for
+        .. highlight:: python
+        .. code-block:: python
+
+            method(m, *args, **kwargs)
+
+        if in a transaction body.
```
tilk: What is the rationale for this behavior?

tilk (approved):
Looks like a straightforward API extension which we probably won't regret later.

PR #203, opened 2026-07-22
Type-preserving Mux and SwitchValue

tilk:
`SwitchValue` and `Mux` in Amaranth always return Values, which isn't good when working with higher level representations. But they are convenient! This PR adds `switch_value` and `mux`, which preserve data, enums and other `ShapeCastable`s.

TODO:

* [x] Look into better stubs for enums, again


transactron/utils/amaranth_ext/functions.py
```diff
@@ -197,6 +204,75 @@ def max_value(*values: ValueBundle) -> Value:
...
+) -> ValueLike:
+    src_loc = get_src_loc(src_loc)
+    cases = list(cases)
+    case_shapes = [shape_of(val) for _, val in cases]
+    shapecastable_shapes = [shape for shape in case_shapes if isinstance(shape, ShapeCastable)]
+    if shapecastable_shapes:
+        shape = cast(ShapeCastable, shapecastable_shapes[0])
+        if any(case_shape != shape for case_shape in shapecastable_shapes):
+            raise ValueError("Different ShapeCastables for different shapes")
+        return shape(SwitchValue(test, [(key, Value.cast(val)) for key, val in cases], src_loc=src_loc))
```
qbojj: I think it should be `shape(SwitchValue(test, [(key, shape(val)) for key, val in cases], src_loc=src_loc))`, as shapes can define conversions.
For example assuming fixed point signal `fixed`:
```py
mux(sig, fixed, 4)
```
should first covert the 4 into the fixed point format (with bit representation e.g `4 << 8`), than mux

qbojj (changes requested):

transactron/utils/amaranth_ext/functions.py
tilk: Huh, are you certain of that? The documentation of Amaranth says this about `ShapeCastable.__call__`:

> For any compliant `obj`, the following condition must hold:
> ```Value.cast(self(obj)) == Value.cast(obj)```

I have a problem with the definition of "compliant":
> This method is called by the Amaranth language to lift `obj`, which may be any value-like object whose shape equals `Shape.cast(self)`, to a higher-level representation, which may be any value-like object with the same shape.

Ints are value-like, but don't have a shape (until casted to a value).

qbojj: Oh yeah. I think all representation convertions are either done at `eq` or when using `shape.const` (that is what amaranth-lang/amaranth#1578 does).
As it is required for all shapes to be the same, than something like may work:
```py
def unify(v):
   return Value.cast(v if isinstance(v, Value) else shape.const(v))
return shape(SwitchValue(test, [(key, unify(val)) for key, val in cases], src_loc=src_loc))
```

tilk: Your code is ignoring ValueCastables.

qbojj: Than simply
```py
def unify(v):
   return Value.cast(v) if isinstance(v, (Value, ValueCastable)) else Value.cast(shape.const(v))
return shape(SwitchValue(test, [(key, unify(val)) for key, val in cases], src_loc=src_loc))
```

qbojj (approved):

Issue #219, opened 2026-08-08
OneHotMux.create wires its inputs with m.d.comb


xThaid:
Under an m.If (or inside any method/transaction body) every input and select bit gets muxed behind the enclosing condition. That's pure overhead: we create wires a private submodule exactly once, so the condition carries no information.

We sadly can't use m.d.top_comb there because the module passed to OneHotMux doesn't have to be a TModule. One idea is to remove `OneHotMux.create` and replace all usages with `one_hot_mux` from functions.py instead.

PR #221, opened 2026-08-10
make one_hot_mux ShapeCastable aware

qbojj:
Makes `one_hot_mux` shape aware - line `mux` and `switch_value`.
~Also fixes the bug with `mux` and `switch_value` where they failed on 0-wide `ValueCastable`s - `SwitchValue` returns 1-wide value when all it's arguments are 0-wide.~

Related to #219.

Transactron uses internally `one_hot_mux` for method argument combiners, thus testing it quite well on many different shape castables.

### DIFF
--- a/transactron/core/body.py
+++ b/transactron/core/body.py
@@ -120,7 +120,7 @@ class Body(TransactionBase["Body"]):
     def _default_combiner(shape: ShapeLike):
         def impl(m: Module, args: Sequence[MethodStruct], runs: Value) -> AssignArg:
             arg = Signal(shape)
-            m.d.comb += arg.eq(one_hot_mux([(runs[i], args[i]) for i in range(len(args))], assert_one_hot=False))
+            m.d.comb += arg.eq(one_hot_mux([(runs[i], args[i]) for i in range(len(args))]))
             return arg
 
         return impl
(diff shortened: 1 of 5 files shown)


transactron/utils/amaranth_ext/functions.py
```diff
@@ -165,6 +165,43 @@ def const_of(value: int, shape: ShapeLike) -> Any:
...
+
+@overload
+def _uniformize_values(
+    values: Iterable[ValueLike],
+) -> tuple[ShapeCastable | None, list[Value]]: ...
+
+
+def _uniformize_values(
+    values: Iterable[ValueLike],
+) -> tuple[ShapeCastable | None, list[Value]]:
```
tilk: How about:
```suggestion
def _uniformize_values(
    values: Iterable[ValueLike],
) -> tuple[Callable[[ValueLike], ValueLike], list[Value]]:
```
And then, in the case for `ValueCastable`, return `Callable[[ValueLike], T]`. Basically return a cast function. You would then avoid case analysis at use sites.

transactron/core/body.py
```diff
@@ -120,7 +120,7 @@ def peek() -> Optional["Body"]:
     def _default_combiner(shape: ShapeLike):
         def impl(m: Module, args: Sequence[MethodStruct], runs: Value) -> AssignArg:
             arg = Signal(shape)
-            m.d.comb += arg.eq(one_hot_mux(runs, args, assert_one_hot=False))
+            m.d.comb += arg.eq(one_hot_mux([(runs[i], args[i]) for i in range(len(args))]))
```
tilk, line 123: Would prefer a `zip`. Does the typechecker complain? For some reason Amaranth's `Value` does not have an `__iter__` but I think it should.

qbojj: typechecker does complain about this

tilk (approved):

PR #223, opened 2026-08-12
Introduce bitmagic helper functions

qbojj:
Closes #222 

Introduce helpers for common bit manipulation tricks to increase readability.

We need to be consistent about naming the naming of those instruction.
For now I treat the values as bit-vectors with LSB at the front (like indexing in amaranth treats them). This may introduce some confusion, as `count_{leading,trailing}_zeros` treat them as bit-vectors with MSB at the front (as how numbers are written down).

### DIFF
--- a/transactron/utils/amaranth_ext/functions.py
+++ b/transactron/utils/amaranth_ext/functions.py
@@ -377,7 +377,7 @@ def one_hot_mux(
             src_loc=1,
         )
 
-    select_first = (select & (~select + 1))[: len(inputs)]
+    select_first = extract_lowest_set_bit(select)
     select_one_hot = select_first if priority else select
 
     if not priority and assert_one_hot:
--- a/transactron/utils/amaranth_ext/functions.py
+++ b/transactron/utils/amaranth_ext/functions.py
@@ -33,6 +33,13 @@ __all__ = [
     "switch_value",
     "mux",
     "one_hot_mux",
+    "extract_lowest_set_bit",
+    "extract_highest_set_bit",
+    "clear_lowest_set_bit",
+    "mask_from_first_set_bit",
+    "mask_after_first_set_bit",
+    "mask_until_first_set_bit",
+    "mask_before_first_set_bit",
 ]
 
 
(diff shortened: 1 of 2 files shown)


transactron/utils/amaranth_ext/functions.py
```diff
@@ -388,3 +394,57 @@ def one_hot_mux(
         return shape_cast(all_data[0])
 
     return shape_cast(or_value([Mux(all_sel[i], all_data[i], C(0, 0)) for i in range(len(all_data))]))
+
+
+def extract_lowest_set_bit(value: Value) -> Value:
```
tilk, line 399: What about the signedness of the result shape?

I think for some of these functions, an argument could be made for different choices:
- `extract_lowest_set_bit`: unsigned, as the result should have at most one set bit and we don't want to smear MSB.
- `clear_lowest_set_bit`: preserve signedness, as the result is basically the same it was modulo one bit.
- `mask_from_first_set_bit`: signed, as we do want the smearing - the mask logically extends forever!
- `mask_until_first_set_bit`, `mask_before_first_set_bit`: unsigned - logically, the zeros extend forever here.
- `mask_after_first_set_bit`: probably signed, but it's an ugly duckling here in a way: for a 1 on a most significant bit, the extension behavior will be wrong.

Edit: This is kind of nitpicking, as I wonder if bit extension is even useful for functions like this. Maybe "all unsigned" is a better pragmatic choice?

qbojj: - `clear_lowest_set_bit`: I'm not sure if preserving signedness is correct, as if we imagine that the signed values < 0 have infinite chain of leading 1s, than `clear_lowest_set_bit(MIN_INT_FOR_WIDTH)` would clear all of them (aka not only the lowest) - correct behaviour requires extension of the result by one bit.
- `mask_after_first_set_bit`: for correct behaviour would require result to be extended by one bit.
- `mask_from_first_set_bit`: for 0-wide `Value`s we would need to output 0-wide signed value - needs extension to 1 bit.
- `mask_until_first_set_bit`, `mask_before_first_set_bit`: effectively return -1 for input=0 -> also would need to be signed with additional bit

I'm not sure if we would rather have correct width preservation or it to behave well when extending the values.

I think the cleanest solution would be strict width (non arithmetic) signals (`BitVector` shape or something like that)

transactron/utils/amaranth_ext/functions.py
```diff
@@ -388,3 +394,57 @@ def one_hot_mux(
...
+    """
+    Extracts the least significant set bit from the input value.
+    If no bits are set, returns 0.
+    For example 0b010100 -> 0b000100.
+    Same as: (1 << count_trailing_zeros(value))[: len(value)].
+    """
```
tilk, line 405: Could you format docstrings properly, so that they render well? In particular, code citations should use double backticks.

tilk (approved):

PR #224, opened 2026-08-12
Make hypothesis even better

tilk:
It turns out that, indeed, `st.data` works just fine with `pysim`, which opens the way for easy use of Hypothesis even for complex tests.

This PR adds:
- Geometric distribution strategy.
- Hypothesis-driven random delays.
- A trick to avoid manually using `wrap_testing_env_next` (without it, hypothesis tests fail with a weird message).
- A CI hypothesis profile which removes time limits (cycle counts are hardware-independent) and turns on derandomization (constant seed).

Stack test is converted to Hypothesis to showcase that dynamic generation works fine.

TODO:
- [x] Draw labels, for cleaner reports.
- [x] Docstrings.
- [x] Tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,7 +34,7 @@ jobs:
           pip3 install --group test .[testing]
 
       - name: Run tests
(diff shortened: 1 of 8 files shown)


qbojj (approved):
Nice

PR #232, opened 2026-08-20
Add to_signal and top_module

### LINKED ISSUE #219: OneHotMux.create wires its inputs with m.d.comb

tilk:
To help with issues like the recent one in #230, this PR introduces a low-friction way to assign values into signals.

Also introduces a way to do top level assignment for ModuleLikes.

Fixes #219.

### DIFF
--- a/transactron/core/tmodule.py
+++ b/transactron/core/tmodule.py
@@ -5,6 +5,7 @@ from amaranth_types import StatementLike, ModuleLike, ValueLike, SwitchKey
 from typing import Optional, Self, NoReturn
 from contextlib import contextmanager
 from amaranth.hdl._dsl import FSM, _guardedcontextmanager
+from transactron.utils.amaranth_ext import top_module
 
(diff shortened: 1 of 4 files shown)


qbojj (approved):

transactron/core/tmodule.py
qbojj: this could now use `top_module(self._m).d["comb"]`, but that may be of limited use (`_top_module` would not show up on the VCD dumps when `top_comb` was not used)

qbojj (approved):

PR #235, opened 2026-08-21
Make utils functions 0-wide signal aware

### LINKED ISSUE #234: popcount on 0-wide values
`popcount` doesn't work on 0-wide values

qbojj:
closes #234

- makes `popcount` 0-wide value aware
- makes `count_{leading,trailing}_zeros` 0-wide value aware
- cleans up `count_{leading,trailing}_zeros` to remove the padding logic

### DIFF
--- a/test/utils/test_utils.py
+++ b/test/utils/test_utils.py
@@ -74,7 +74,7 @@ class PopcountTestCircuit(Elaboratable):
         return m
 
 
-@pytest.mark.parametrize("size", [2, 3, 4, 5, 6, 8, 10, 16, 21, 32, 33, 64, 1025])
+@pytest.mark.parametrize("size", [0, 2, 3, 4, 5, 6, 8, 10, 16, 21, 32, 33, 64, 1025])
 class TestPopcount(TestCaseWithSimulator):
     @pytest.fixture(scope="function", autouse=True)
     def setup_fixture(self, size):
--- a/test/utils/test_utils.py
+++ b/test/utils/test_utils.py
@@ -18,8 +18,7 @@ from transactron.utils import (
     mask_until_first_set_bit,
     mask_before_first_set_bit,
 )
-from transactron.testing.input_generation import shrinkable_lists
-from hypothesis import given, settings
+from hypothesis import given
 import hypothesis.strategies as st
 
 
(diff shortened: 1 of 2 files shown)


transactron/utils/amaranth_ext/functions.py
```diff
@@ -91,27 +91,21 @@ def iter(s: Value, step: int) -> Value:
...
-    zeros_prepend_count = closest_pow_2_of_s - slen
-    value = iter(Cat(C(0, shape=zeros_prepend_count), s), slen_log)
+    value = iter(s.as_unsigned(), ceil_log2(slen))
 
     # 0 number edge case
     # if s == 0 then iter() returns value off by 1
     # this switch negates this effect
-    result = Mux(s.any(), value, slen)
-    return result
+    return Mux(s.any(), value, slen)
```
qbojj: There is also an option to implement this as:
```py
def count_trailing_zeros(s: Value) -> Value:
    if len(s) == 0:
        return C(0, 0)

    def iter(s: Value, step: int) -> Value:
        # if no bits left - return empty value
        if step == 0:
            return C(0, 0)

        # boudaries of upper and lower halfs of the value
        partition = 2 ** (step - 1)

        if len(s) < partition:
            return Cat(iter(s, step - 1), 0)

        # recursive call
        upper_value = iter(s[partition:], step - 1)
        lower_value = iter(s[:partition], step - 1)

        # if there are lit bits in lowerhalf - take result directly from recursive value
        # otherwise add 1 << (step - 1) to upper value and return
        return Mux(s[:partition].any(), Cat(lower_value, 0), Cat(upper_value, 1))

    return iter(s.as_unsigned(), ceil_log2(len(s) + 1))
```

to also remove the special case for s == 0 (this also simplifies that case for non POT lengths)

test/utils/test_utils.py
tilk: Can CLZ and CTZ tests also test the size=0 case?

transactron/utils/amaranth_ext/functions.py
tilk: Any downsides to this? Looks better.

tilk:
Cool test refactor!

tilk (approved):

PR #236, opened 2026-08-21
Option type

tilk:
I'm getting sick of bare `valid` bits. This is an attempt at something nicer.

TODO:
* [x] amaranth-stubs fix
* [ ] Docstrings
* [ ] Tests

In future: look in Transactron APIs for valid bits and replace with this.


tilk:
Unexpectedly hit an Amaranth issue: https://github.com/amaranth-lang/amaranth/pull/1678

PR #239, opened 2026-08-22
remove undriven signal detection

### LINKED ISSUE #238: Inaccurate undriven signal detection
`VerilogDebugWrapper` finds all statements that assign to signals, to later force their synthesis by assigning the initial value in the comb domain. This is not enough, as some signals are primitive (either IO to the `Instance`s or `Memory`) and are never the LHS of a statement.

We should probably leave original signals alone and always generate a new signal with copied value.

qbojj:
closes #238


tilk (approved):

tilk:
Cool trick!

